PR #2, opened 2015-03-23
Updated godoc comments.

cmars:
Just some minor godoc improvements, as recommended in http://blog.golang.org/godoc-documenting-go-code.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -269,8 +269,7 @@ func (d Decimal) Div(d2 Decimal) Decimal {
 	return ret
 }
 
-// Cmp compares d and d2 and returns -1, 0 or 1.
-// The smallest scale of d and d2 are used in comparison.
+// Cmp compares the numbers represented by d and d2 and returns:
 //
 //     -1 if d <  d2
 //      0 if d == d2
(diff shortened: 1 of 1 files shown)


ox:
This looks great, thanks! I'll let @vadimg give it a 1-over before pushing the button.

decimal.go
```diff
@@ -279,20 +284,23 @@ func (d Decimal) Cmp(d2 Decimal) int {
 	return rd.value.Cmp(rd2.value)
 }
 
+// Equals returns whether d == d2 at the smallest scale among d and d2.
```
vadimg: This comment is confusing, just say that it returns whether the numbers represented by d and d2 are equal, regardless of their internal decimal representation.

decimal.go
```diff
@@ -265,11 +269,12 @@ func (d Decimal) Div(d2 Decimal) Decimal {
 	return ret
 }
 
-// Cmp compares x and y and returns -1, 0 or 1
+// Cmp compares d and d2 and returns -1, 0 or 1.
+// The smallest scale of d and d2 are used in comparison.
```
vadimg: This comment is also confusing, just say it compares the numbers represented by the decimals, regardless of what their internal representation is.

vadimg:
Looks good, thanks! Just a comment on two of your comments...

cmars:
Updated, PTAL

vadimg:
Thanks for adding those comments!

PR #4, opened 2015-03-24
Added NewFromRat implementation.

tasdomas:
It would be convenient if decimal supported initialization from *big.Rat as well.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -158,6 +158,18 @@ func NewFromFloatWithExponent(value float64, exp int32) Decimal {
 	}
 }
 
+// NewFromRat returns a decimal from a big.Rat representation.
+func NewFromRat(value *big.Rat) Decimal {
+	// fast path, where Rat is an int
+	if value.IsInt() {
+		f, _ := value.Float64()
+		return New(int64(math.Floor(f)), 0)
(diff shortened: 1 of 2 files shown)


decimal.go
```diff
@@ -158,6 +158,18 @@ func NewFromFloatWithExponent(value float64, exp int32) Decimal {
 	}
 }
 
+// NewFromRat returns a decimal from a big.Rat representation.
+func NewFromRat(value *big.Rat) Decimal {
+	// fast path, where Rat is an int
+	if value.IsInt() {
+		f, _ := value.Float64()
```
vadimg: this only works for small Rats. big.Rats won't necessarily fit into a Float64, so you'll be losing information here.

decimal.go
```diff
@@ -158,6 +158,18 @@ func NewFromFloatWithExponent(value float64, exp int32) Decimal {
...
+		return New(int64(math.Floor(f)), 0)
+	}
+
+	floatValue, _ := value.Float64()
```
vadimg: same thing here. You can't fit every big.Rat into 64 bits

tasdomas:
Yeah, you're right.

PR #6, opened 2015-06-10
Added methods Min and Max to Decimal.

gjvnq:


decimal.go
```diff
@@ -447,6 +447,26 @@ func (d Decimal) MarshalText() (text []byte, err error) {
 	return []byte(d.String()), nil
 }
 
+func (d Decimal) Min(others ...Decimal) Decimal {
```
vadimg: I would prefer these to be functions, not methods. It looks much better to do `decimal.Min(a, b, c)` than to do `a.Min(b, c)`

decimal_test.go
```diff
@@ -449,6 +449,26 @@ func TestDecimal_Cmp2(t *testing.T) {
...
+		t.Errorf("Error")
+	}
+}
+
+func TestDecimal_Max(t *testing.T) {
+	a := NewFromFloat(3.14)
+	b := NewFromFloat(1.6)
+	c := NewFromFloat(-2)
+
+	if a.Max(b, c) != a {
```
vadimg: You shouldn't use == or != to test equality of Decimals, you should use .Equals

decimal.go
gjvnq: I've fixed that on commit 65c6084.

vadimg:
This code is not up to the standards of this library. For example, here are some issues:
- What happens when it's called with no arguments?
- The tests are not thorough enough
- There's no documentation
- Why do you check for <= and >= instead of < and >?

I decided it would be easier and faster for me to just implement this myself.

PR #10, opened 2015-07-24
Add scientific notation parsing

thiagoarruda:
Here's the code from the Gist with some tests added. I'm not sure how exhaustive they have to be, so I created one for each common scenario, and a few for invalid formats.

_Note: this is my first time submitting a pull request on Github, please let me know if I've done anything wrong, or if you need me to change/add something._

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -81,17 +81,17 @@ func New(value int64, exp int32) Decimal {
 //
 func NewFromString(value string) (Decimal, error) {
 	var intString string
-	var exp int32
+	var exp int64
 
 	// Check if number is using scientific notation
 	eIndex := strings.IndexAny(value, "Ee")
 	if eIndex != -1 {
-		expInt, err := strconv.ParseInt(value[eIndex+1:len(value)], 10, 32)
+		expInt, err := strconv.ParseInt(value[eIndex+1:len(value)], 10, 64)
 		if err != nil {
 			return Decimal{}, fmt.Errorf("can't convert %s to decimal: exponent is not numeric", value)
 		}
 		value = value[0:eIndex]
-		exp = int32(expInt)
+		exp = expInt
 	}
 
 	parts := strings.Split(value, ".")
@@ -102,11 +102,7 @@ func NewFromString(value string) (Decimal, error) {
 	} else if len(parts) == 2 {
 		intString = parts[0] + parts[1]
 		expInt := -len(parts[1])
-		if expInt < math.MinInt32 {
-			// NOTE(vadim): I doubt a string could realistically be this long
-			return Decimal{}, fmt.Errorf("can't convert %s to decimal: fractional part too long", value)
-		}
-		exp += int32(expInt)
+		exp += int64(expInt)
 	} else {
 		return Decimal{}, fmt.Errorf("can't convert %s to decimal: too many .s", value)
 	}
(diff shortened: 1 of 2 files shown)


decimal_test.go
```diff
@@ -32,6 +32,16 @@ var testTable = map[float64]string{
 	.1000000000000008:   "0.1000000000000008",
 }
 
+var testTableScientificNotation = map[string]string{
+	"1e9":       "1000000000",
+	"2.41E-3":   "0.00241",
+	"24.2E-4":   "0.00242",
+	"243E-5":    "0.00243",
+	"-1e-5":     "-0.00001",
```
vadimg: You should not test negatives directly, you should add negative versions of all of these to testTableScientificNotation in the init() function

decimal_test.go
```diff
@@ -32,6 +32,16 @@ var testTable = map[float64]string{
...
+	"-245E3":    "-245000",
+	"1.2345E-1": "0.12345",
+}
```
vadimg: Some more test cases you should add:
0e5
0e-5
123.456e0
123.456e2
123.456e10

decimal_test.go
```diff
@@ -95,6 +116,9 @@ func TestNewFromStringErrs(t *testing.T) {
 		".5.2",
 		"8..2",
 		"8.1.",
+		"1e",
+		"1-e",
+		"1e9e",
```
vadimg: Some more test cases you should add:
1ee9
1ee
1e1.2
123.456e1.3
1e-1.2
123.456e-1.3

decimal.go
```diff
@@ -95,7 +106,7 @@ func NewFromString(value string) (Decimal, error) {
 			// NOTE(vadim): I doubt a string could realistically be this long
 			return Decimal{}, fmt.Errorf("can't convert %s to decimal: fractional part too long", value)
 		}
-		exp = int32(expInt)
+		exp += int32(expInt)
```
vadimg: Since you're adding `expInt` to `exp` here (both int32s), there could potentially be an int32 overflow. I think a good way to handle this would be to make `exp` an int64, and then right before you create and return the Decimal, to return an error if it is < math.MinInt32 or > math.MaxInt32.

You can then remove the the `if expInt < math.MinInt32` above this line.

decimal.go
```diff
@@ -81,21 +81,28 @@ func New(value int64, exp int32) Decimal {
 //
 func NewFromString(value string) (Decimal, error) {
 	var intString string
-	var exp int32
+	var exp int64
+
+	// Check if number is using scientific notation
+	eIndex := strings.IndexAny(value, "Ee")
+	if eIndex != -1 {
+		expInt, err := strconv.ParseInt(value[eIndex+1:len(value)], 10, 64)
```
vadimg: You don't need the `len(value)` in the slice bounds: http://play.golang.org/p/v4cw0g_TkK

decimal.go
```diff
@@ -81,21 +81,28 @@ func New(value int64, exp int32) Decimal {
...
+		if err != nil {
+			return Decimal{}, fmt.Errorf("can't convert %s to decimal: exponent is not numeric", value)
+		}
+		value = value[0:eIndex]
```
vadimg: Similarly, you don't need the 0 here: http://play.golang.org/p/HihfmRUlax

vadimg:
This looks really good, just fix a few small style nits, and I'll merge it. Thank you for adding this feature!

decimal.go
vadimg: Agh, I just noticed this: if `expInt == math.MinInt64`, the line where you `exp += int64(expInt)` will cause an integer overflow. You should parse this as a 32 bit int, and cast it to an int64 when you assign to `exp` below.

decimal_test.go
```diff
@@ -95,6 +126,15 @@ func TestNewFromStringErrs(t *testing.T) {
...
+		"1ee9",
+		"1ee",
+		"1e1.2",
+		"123.456e1.3",
+		"1e-1.2",
+		"123.456e-1.3",
```
vadimg: You should also add this test `"123.456e" + strconv.FormatInt(math.MinInt64, 10)` here

vadimg:
Ah yes, you are right. I will fix this.

PR #12, opened 2015-09-03
replaced divsion code

raverkamp:
Hello,
I replaced the division code for decimals. Now only divsion with remainder for big.Int is used.
I added:
the function DivRound which takes the precision as a additionall argument.
the function QuoRem which has the same semantic as big.Int.QuoRem but with precision as additional argument.
some tests 
benchmarks to make sure that I did not mess up performance.
Kind regards,
Roland Averkamp

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -287,6 +287,70 @@ func (d Decimal) Div(d2 Decimal) Decimal {
 	return ret
 }
 
+// return quotient q and remainder r such that
+// scale := -precision
+// d = d2 * q + r such that q an integral multiple of 10^scale
+//  and 0 <= r < abs(d2) * 10 ^scale
+//  this holds if d>=0
+// if d <0  then r =<0 with r>= -abs(d2) * 10 ^ scale
+func (d Decimal) QuoRem(d2 Decimal, precision int32) (Decimal, Decimal) {
+	scale := -precision
+	e := int64(d.exp - d2.exp - scale)
+	var aa, bb, expo big.Int
+	var scalerest int32
+	// d = a 10^ea
+	// d2 = b 10^eb
+	if e < 0 {
+		aa = *d.value
+		expo.SetInt64(-e)
+		bb.Exp(tenInt, &expo, nil)
+		bb.Mul(d2.value, &bb)
+		scalerest = d.exp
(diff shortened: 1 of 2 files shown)


victorquinn:
Hey @raverkamp , I am just taking over as maintainer of this repo and saw this big change.

First, apologies this went so long without attention!

My question is: What benefit does this provide over the old method? 

Aside from removing the dependency on big.NewRat's Quo method, it looks like it accomplishes the same thing, so I'm wondering why would it make sense to add all this additional complexity to this library when we can leverage the big library to do that for us? 

It seems to me this adds a lot of error prone code to this library without providing much benefit. the [Rat library is very extensively tested](https://github.com/golang/go/blob/master/src/math/big/rat_test.go) and more widely used than this library so leveraging that test suite and large userbase seems like a big win for this library.

If this change made this library better, faster, more accurate, or added new features it would make sense, but it seems like it does none of the above and adds complexity. Am I missing anything?

raverkamp:
Hello Victor,
the current code for division is:
```
// NOTE(vadim): division is hard, use Rat to do it
	ratNum := d.Rat()
	ratDenom := d2.Rat()

	quoRat := big.NewRat(0, 1).Quo(ratNum, ratDenom)

	// HACK(vadim): converting from Rat to Decimal inefficiently for now
	ret, err := NewFromString(quoRat.FloatString(DivisionPrecision))
```
Most likely NewRat.Quo only removes the GCD. The heavy lifting is done by  FloatString.

My main motivation is the comment:
// HACK(vadim): converting from Rat to Decimal inefficiently for now

The new code is about 3 times faster:
`[roland@averkamp1 decimal]$ go test -test.bench ".*" -test.benchmem
Benchmark_DivideOriginal-4   	       1	7865926206 ns/op	1647044416 B/op	39474820 allocs/op
Benchmark_DivideNew-4        	       1	2062495137 ns/op	600448640 B/op	12996666 allocs/op
Benchmark_Cmp-4              	       2	 564685551 ns/op	      32 B/op	       1 allocs/op
`

victorquinn:
Thanks for the clarification @raverkamp that compelling! 

Thanks again for your work and apologies it took this long to get it merged!

PR #14, opened 2015-11-23
Added support for Numeric type in db storage

faide:
Hi,

I would like to propose this small PR that enables to store data as Numeric in the database.

I tested it with a small attached self contained program: 
https://gist.github.com/faide/3db008dced89b3872e54

If run against actual trunk it will fail, but when run against the patched version if will work as expected.

I am open to discussions and/or suggestions on how to unittest this code or generally make it better :)

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -481,7 +481,7 @@ func (d *Decimal) Scan(value interface{}) error {
 	case int64:
 		// at least in sqlite3 when the value is 0 in db, the data is sent
 		// to us as an int64 instead of a float64 ...
-		*d = NewFromFloat(float64(value.(int64)))
+		*d = New(value.(int64), 0)
 		return err
 
 	default:
--- a/decimal.go
+++ b/decimal.go
@@ -471,22 +471,22 @@ func (d Decimal) MarshalJSON() ([]byte, error) {
 func (d *Decimal) Scan(value interface{}) error {
 	var err error
 	// first try to see if the data is stored in database as a Numeric datatype
-	switch value.(type) {
+	switch v := value.(type) {
 
 	case float64:
 		// numeric in sqlite3 sends us float64
-		*d = NewFromFloat(value.(float64))
+		*d = NewFromFloat(v)
 		return err
 
 	case int64:
 		// at least in sqlite3 when the value is 0 in db, the data is sent
 		// to us as an int64 instead of a float64 ...
-		*d = New(value.(int64), 0)
+		*d = New(v, 0)
 		return err
 
 	default:
 		// default is trying to interpret value stored as string
-		str, err := unquoteIfQuoted(value)
+		str, err := unquoteIfQuoted(v)
 		if err != nil {
 			return err
 		}
--- a/decimal.go
+++ b/decimal.go
@@ -469,6 +469,14 @@ func (d Decimal) MarshalJSON() ([]byte, error) {
 
 // Scan implements the sql.Scanner interface for database deserialization.
 func (d *Decimal) Scan(value interface{}) error {
+	var err error
+	// first try to see if the data is stored in database as a Numeric datatype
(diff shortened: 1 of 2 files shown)


decimal.go
```diff
@@ -469,13 +469,30 @@ func (d Decimal) MarshalJSON() ([]byte, error) {
...
+
+	case float64:
+		// numeric in sqlite3 sends us float64
+		*d = NewFromFloat(value.(float64))
+		return err
+
+	case int64:
+		// at least in sqlite3 when the value is 0 in db, the data is sent
+		// to us as an int64 instead of a float64 ...
+		*d = NewFromFloat(float64(value.(int64)))
```
vadimg: This will lose precision if the value is indeed an int64. float64 only has 52 bits of precision, so it can't exactly represent every 64 bit integer. You should create the decimal directly from the int64, by calling `New(value, 0)`.

faide:
I updated the PR to avoid going through float64 if we have an int64. (this is only the case when dealing with zeros)

faide:
Do you think the small code snippet I provided as a test case could be integrated in the test suite after some adaptations?

faide:
I added the test code (with updated test for zero values) here:
https://github.com/faide/dectest

decimal.go
```diff
@@ -469,13 +469,30 @@ func (d Decimal) MarshalJSON() ([]byte, error) {
 
 // Scan implements the sql.Scanner interface for database deserialization.
 func (d *Decimal) Scan(value interface{}) error {
-	str, err := unquoteIfQuoted(value)
-	if err != nil {
+	var err error
+	// first try to see if the data is stored in database as a Numeric datatype
+	switch value.(type) {
```
vadimg: So you don't have to do type assertions (value.(float64), etc), you should assign the type switch to t, like in here: https://golang.org/doc/effective_go.html#type_switch

decimal.go
```diff
@@ -469,13 +469,30 @@ func (d Decimal) MarshalJSON() ([]byte, error) {
 
 // Scan implements the sql.Scanner interface for database deserialization.
 func (d *Decimal) Scan(value interface{}) error {
-	str, err := unquoteIfQuoted(value)
-	if err != nil {
+	var err error
```
vadimg: why create this err?

decimal.go
```diff
@@ -469,13 +469,30 @@ func (d Decimal) MarshalJSON() ([]byte, error) {
...
+	// first try to see if the data is stored in database as a Numeric datatype
+	switch value.(type) {
+
+	case float64:
+		// numeric in sqlite3 sends us float64
+		*d = NewFromFloat(value.(float64))
+		return err
```
vadimg: you should return nil here, instead of the empty err above. Returning the err makes me have to find the definition of the err and make sure that it was never set to anything in order to parse this code.

decimal.go
```diff
@@ -469,13 +469,30 @@ func (d Decimal) MarshalJSON() ([]byte, error) {
...
+
+	case int64:
+		// at least in sqlite3 when the value is 0 in db, the data is sent
+		// to us as an int64 instead of a float64 ...
+		*d = New(value.(int64), 0)
 		return err
```
vadimg: same here

vadimg: same here

vadimg:
It's hard to add a test, because it depends on a sqlite library, which doesn't come with go. Therefore adding a sqlite test would require sqlite as a dependency... I wonder how other libraries handle this?

Go's own sql library creates a mock driver: https://golang.org/src/database/sql/fakedb_test.go but unfortunately it is private.

faide:
Vadim,

I took your remarks and changed the code, I also proposed in the PR a unit test that uses and validates the Scan method.

Would you mind commenting on the test and new code :)

vadimg:
Looks good, thank you!

PR #15, opened 2015-12-11
Add a fast track to d.Cmp(d2) when d.exp == d2.exp

pwiecz:
This makes a noticeable difference when one is sorting a large slice
of decimals when most of them have the same precision, as it is saving
a lot of calls to rescale().


vadimg:
lgtm! thank you!

PR #16, opened 2015-12-24
Make Value take a pointer receiver

tomlinford:
The `Value` method needs to take a pointer receiver so that null values can be inserted into the database with ease.

This PR shouldn't affect any existing behavior since the arguments to `Exec` get converted into an `interface{}`.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -495,7 +495,10 @@ func (d *Decimal) Scan(value interface{}) error {
 }
 
 // Value implements the driver.Valuer interface for database serialization.
-func (d Decimal) Value() (driver.Value, error) {
+func (d *Decimal) Value() (driver.Value, error) {
+	if d == nil {
+		return nil, nil
+	}
 	return d.String(), nil
 }
 
--- a/decimal_test.go
+++ b/decimal_test.go
@@ -874,6 +874,27 @@ func TestDecimal_Scan(t *testing.T) {
 	}
 }
 
+func TestDecimal_Value(t *testing.T) {
+	// check that nil is handled appropriately
+	var decimalPtr *Decimal
+	value, err := decimalPtr.Value()
+	if err != nil {
+		t.Errorf("(*Decimal)(<nil>) failed with message: %s", err)
+	} else if value != nil {
+		t.Errorf("%v is not nil", value)
+	}
+
+	// check that normal case is handled appropriately
+	a := New(1234, 2)
+	expected := "123400"
+	value, err = a.Value()
+	if err != nil {
+		t.Errorf("Decimal(123400).Value() failed with message: %s", err)
+	} else if value.(string) != expected {
+		t.Errorf("%s does not equal to %s", a, expected)
+	}
+}
+
 // old tests after this line
 
 func TestDecimal_Scale(t *testing.T) {


manoharank:
@vadimg Any chance to merge this PR? We need to store null values in a decimal column.

manoharank:
@tomlinford What about handling `nil` in Scan function?

tomdionysus:
:+1: This would be really handy for us as well

jayd3e:
Definitely need this as well :).  Specifically what @manoharank mentioned.  The `nil` value is not covered in the Scan function.

snikch:
This was a breaking change for our implementation. You now cannot use a `decimal.Decimal` as a parameter for database inserts, it has to be the pointer value.

```go
package main

import (
	"database/sql"
	"fmt"

	_ "github.com/go-sql-driver/mysql"
)

func main() {
	query := "SELECT ?"
	db, err := sql.Open("mysql", "polecat:polecat@tcp(127.0.0.1:33307)/")
	if err != nil {
		panic(err)
	}

	_, err = db.Exec(query, &decimal.Zero)
	if err != nil {
		fmt.Printf("Failed to use with ptr: %s\n", err.Error())
	}
	_, err = db.Exec(query, decimal.Zero)
	if err != nil {
		fmt.Printf("Failed to use without ptr: %s\n", err.Error())
	}
}
```

```
Failed to use without ptr: sql: converting Exec argument #0's type: unsupported type decimal.Decimal, a struct
```

jayd3e:
@snikch that's odd, b/c decimal.Decimal.Value() should always return a string in the case that decimal.Zero is passed in.  Does removing the addition of the `if nil { return nil, nil }` check fix your issue?

jayd3e:
Also, this might be a stupid question, but do you know if github.com/go-sql-driver uses the `Valuer` interface when parsing arguments to `Exec`?  I know pgx only just implemented it a year ago.

snikch:
Even though you can call `.Value` on it, because of the implicit address taking that allows you to call a pointer method on a non pointer variable, it's not able to pass the `x.(driver.Valuer)` test.

```go
	var without interface{} = decimal.Zero
	var with interface{} = &decimal.Zero
	if _, ok := with.(driver.Valuer); !ok {
		fmt.Printf("with is not considered a driver.Valuer")
	}
	if _, ok := without.(driver.Valuer); !ok {
		fmt.Printf("without is not considered a driver.Valuer")
	}
```

```
without is not considered a driver.Valuer
```

snikch:
Which is what it does internally https://golang.org/src/database/sql/driver/types.go?s=5015:5047#L216

jayd3e:
I'm really confused by this.  According to the spec https://golang.org/ref/spec#Type_assertions, they don't mention pointers as a unique case.  Usually in Go, pointers are derefereced automatically, so passing the value vs. a pointer should have no distinction.  Furthermore, all this PR did was add a nil check, which should affect the flow of your program when passing in decimal.Zero.  Are you sure it was this specific PR that caused the issue?

jayd3e:
Shouldn't*

snikch:
The distinction is that the method signature changed from `func (d Decimal) Value() (driver.Value, error)` to `func (d *Decimal) Value() (driver.Value, error)`.

If you read this part of the spec, I think it explains it.

https://golang.org/ref/spec#Method_values

> As with method calls, a reference to a non-interface method with a pointer receiver using an addressable value will automatically take the address of that value: t.Mp is equivalent to (&t).Mp.

The key here is "non-interface method". Since `Value` is an interface method, it will not automatically take the address of the variable.

jayd3e:
Ah thanks for explaining.  And we need `d` to be a pointer b/c we can't compare it to nil otherwise right?

snikch:
Correct! So it's unfortunately a lose lose situation, as I can definitely see the need for having null values. For normal `sql` usage you could always just take the address of the variable holding a `decimal.Decimal` when using it in a query, except we use the popular package `sqlx` and pass structs around, which would mean we'd need to change the struct field types from `decimal.Decimal` to `*decimal.Decimal` which would involve some changes we're not willing to make to our fundamental data types.

I can't think of any solution off the top of my head that satisfies both requirements, without adding some sort of field on the decimal itself, i.e. `IsNull` which is kinda dumb 😢

snikch:
I guess this is why `sql` has the `NullXyx` datatypes, when users want to avoid a pointer.

We could add

```go
type NullDecimal struct {
  Decimal
  Valid bool
}

// Scan implements the Scanner interface.
func (nd *NullDecimal) Scan(value interface{}) error {
  if value == nil {
    nd.Valid = false
    return nil
  }
  nd.Valid = true
  return nd.Decimal.Scan(value)
}

// Value implements the driver Valuer interface.
func (nd NullDecimal) Value() (driver.Value, error) {
  if !nd.Valid {
    return nil, nil
  }
  return nd.Value()
}
```

jayd3e:
Yeah, I think this would work well.  I vote to reverse the previous change in Value/Scan and implement a `NullDecimal` struct that follows the same paradigm as `NullString`.  I like the idea of naming it `Null`, b/c then you would instantiating it like `var nullDecimal decimal.Null`.  But I could get behind being more explicit like `NullString` is.

jayd3e:
It would follow convention much more the second way as well, as you had it.

braindev:
Is there an issue I can follow for reverting the change and adding NullDecimal?

PR #17, opened 2016-01-13
Prevent unnecessary int64 overflow in NewFromFloat

nvanbenschoten:
Previously, a float64 that would overflow an int64 in `NewFromFloat ` would attempt to use the fast path. This is now prevented.


vadimg:
Woah, good catch. Thanks for the pull request!

nvanbenschoten:
No problem, thanks for the library!

PR #20, opened 2016-01-22
Added implementation of DigitsValue() that exposes the big.Int value part of the decimal.

gintas:
The current API exposes the Exponent but not the underlying big.Int(). This caused me some needless pain with constructing a bridge to another implementation of decimals that also uses an integer and an exponent underneath.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -532,6 +532,16 @@ func (d Decimal) MarshalText() (text []byte, err error) {
 	return []byte(d.String()), nil
 }
 
+// GobDecode implements the gob.GobDecoder interface for gob serialization.
+func (d *Decimal) GobDecode(data []byte) error {
+	return d.UnmarshalText(data)
+}
+
+// GobEncode implements the gob.GobEncoder interface for gob serialization.
+func (d Decimal) GobEncode() ([]byte, error) {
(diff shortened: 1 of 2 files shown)


Stefflan00:
what is the status? Need that part as well.

decimal.go
```diff
@@ -313,6 +313,12 @@ func (d Decimal) Equals(d2 Decimal) bool {
 	return d.Cmp(d2) == 0
 }
 
+// DigitsValue returns the integer that represents the digits of the decimal.
+// This is the value of the decimal scaled by 10^Exponent().
+func (d Decimal) DigitsValue() *big.Int {
```
vadimg: Can you rename this to Digits() to be consistent with the Exponent() method?

vadimg: Actually, I thought about it a bit more, and I think it should be called Coefficient().

Coefficient and Exponent are the proper terms for these values. I should have named d.value d.coefficient, but I didn't.

Other proper terms are: significand and mantissa[1]

DigitsValue is not the right name for this.

[1] https://en.wikipedia.org/wiki/Significand

victorquinn:
This was resolved by #42 since this PR seemed to have been abandoned

PR #23, opened 2016-03-11
Gob

gintas:


judwhite:
The `time` packages calls MarshalBinary inside GobEncode ([source](https://golang.org/src/time/time.go#L920)).

Should this package do something similar, or is it not feasible?

klokare:
> The time packages calls MarshalBinary inside GobEncode (source).
> Should this package do something similar, or is it not feasible?

I think so. Separate from this pull request I submitted #49 which provides the encoding.BinaryMarshaler interface methods: MarshalBinary and UnmarshalBinary. That implementation is much faster than this one (I, too, originally solved this by reusing the bytes from MarshalText, see #34 for the performance difference of using the new method in #49). Now that Decimal implements BinaryMarshaler, GOB encoding/decoding should just work: 

> Gob can encode a value of any type implementing the GobEncoder or encoding.BinaryMarshaler interfaces by calling the corresponding method, in that order of preference.
> Gob can decode a value of any type implementing the GobDecoder or encoding.BinaryUnmarshaler interfaces by calling the corresponding method, again in that order of preference.

-from [GOB package docs](https://golang.org/pkg/encoding/gob/)

victorquinn:
GobEncode and GobDecode implemented in ed31524af4041437c0fc9fe6da5f7227d858d2b5 using MarshalBinary and UnmarshalBinary as @klokare suggested so I'm closing this PR.

Also, this PR was changing some things unrelated to GobEncode/GobDecode (DigitsValue) which is why I didn't merge it earlier.

PR #24, opened 2016-05-13
Type switch in 'unquoteIfQuoted'

AdrianLungu:
I've run into a case where, when using PGX with SQLX, the PostgreSQL database seems to be returning the column type 'numeric' as a string which makes 'unquoteIfQuoted' return an error saying it can't convert it to a byte array. 

I've added a switch to check for both []byte and string and also ran the tests to make sure everything's fine this way. All ok!


vadimg:
This functionality has been added by merging https://github.com/shopspring/decimal/pull/27

PR #27, opened 2016-08-03
Handle strings being passed to Scan()

chrismrivera:
lib/pq can now pass strings to Scan() as of:
https://github.com/lib/pq/commit/e2402a7cd1e57e08a576b94cdfed36ae30366545


manuel-huez:
We need that as well, would be awesome if it could be merged.

PR #29, opened 2016-09-19
add Pow() (fix #26)

### LINKED ISSUE #26: Pow implementation
When calculating taxes and converting between day/month/year bases, it is necessary to use Pow with decimal exponent. Is there any way to do this with this library?

martinlindhe:
fix issue #26


victorquinn:
Looks great, thanks for the addition! Apologies it took so long to get reviewed.

PR #30, opened 2016-10-06
Add unbiased rounding algorithm

connor4312:
This adds a "round to even" implementation for unbiased rounding.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -56,6 +56,7 @@ var Zero = New(0, 1)
 
 var zeroInt = big.NewInt(0)
 var oneInt = big.NewInt(1)
+var twoInt = big.NewInt(2)
 var fiveInt = big.NewInt(5)
 var tenInt = big.NewInt(10)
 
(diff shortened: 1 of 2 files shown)


connor4312:
Looks like Travis is set to use 1.4 and below that don't have `t.Run()`. Would you like me to bump the Travis versions or modify my code to avoid using that method?

decimal.go
```diff
@@ -422,6 +423,50 @@ func (d Decimal) Round(places int32) Decimal {
...
+// equals 0.5, this function will round up for odd numbers and down for
+// even numbers. Negative values are treated symmetrically.
+//
+// Example:
+//
+// 	   NewFromFloat(5.5).Round(0).String() // output: "6"
+// 	   NewFromFloat(8.5).Round(0).String() // output: "8"
+//
+func (d Decimal) RoundFair(places int32) Decimal {
+	shift := big.NewInt(abs(int64(places)-int64(d.exp)) - 1)
```
victorquinn: Please update this to use `math.Abs` rather than implementing our own `abs()` function in this library.

For example:
```go
shift := big.NewInt(int64(math.Abs(float64(places) - float64(exp)) - 1))
```

victorquinn:
Thanks for the addition! (and apologies this took so long, I recently took over maintenance of this repo so future changes shouldn't go this long without getting looked at!)

I requested one change in the code, and yes, please try not to use `t.Run()` so we keep backwards compatibility with lower versions of go. This is not super critical but since it's pretty easy and none of the rest of the tests use `t.Run()`, seems pretty easy to keep that compatibility.

alpe:
@connor4312 @victorquinn I run into the rounding issue so thanks for the PR. Any chance to get this fixed and merged soon?

alpe:
I expected `1.444445` with `RoundFair(2)` to be `1.45` but it returns `1.44`. Looks like a 🐛

connor4312:
Hey, yea. This implementation is probably not very good. I tried to find reference implementations of this unbiased rounding implementations, but at the time was unsuccessful. Maybe I'll revisit this as a weekend project.

victorquinn:
Your implementation is fine @connor4312 I would just prefer to use a well tested abs function rather than having our own which is not well tested (silly as that may seem given it's such a trivial easy thing to write). So fix this then I'm happy to merge.

connor4312:
Ok, will do! Perhaps in my naiveness I stumbled upon the correct solution 😄

funny-falcon:
> I expected 1.444445 with RoundFair(2) to be 1.45 but it returns 1.44. Looks like a 🐛

@alpe As far as I understand "round to even", 1.44 is right answer

funny-falcon:
@alpe even more: 1.444445 rounds to 1.44 with any rule. 1.445 rounds to 1.44 with "round to even" and to 1.45 with school "half away from zero".

alpe:
Apologies, it was very late last night, when I ran into the issue. Which is no issue, thanks @funny-falcon

achainan:
@victorquinn can you take me out of this project. Tried to remove my access but couldn't. Thanks.

PR #34, opened 2016-11-04
Add binary (un)marshalling support

klokare:
Implement encoding.BinaryMarshaler and Binary.Unmarshaller. Add tests
to verify.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -548,6 +548,18 @@ func (d Decimal) MarshalText() (text []byte, err error) {
 	return []byte(d.String()), nil
 }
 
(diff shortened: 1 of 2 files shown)


klokare:
Removing this pull request. I will submit a new one that does a better job. Here are the stats. Old is this pull request and New is the one that will be submitted next. The first 2 lines represent running the entire test table. The second two lines is running just 1234.567890123456.

BenchmarkOldBin-4               	   20000	     73760 ns/op
BenchmarkNewBin-4               	  100000	     13354 ns/op

BenchmarkSingleOldBin-4         	 1000000	      1629 ns/op
BenchmarkSingleNewBin-4         	 5000000	       311 ns/op

PR #38, opened 2016-12-12
Add Sign method

edsrzf:
d.Sign() is a more convenient way of saying d.Cmp(Zero).


edsrzf:
Rebased against recent changes.

victorquinn:
Looks great, thanks for the addition!

PR #39, opened 2016-12-28
Handle float32 in Scan method

james-maloney:
Using MySQL version 5.7.16 and github.com/go-sql-driver/mysql the float type is coming back as a float32 which is not currently handled in the Scan method.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -524,10 +524,7 @@ func (d *Decimal) Scan(value interface{}) (err error) {
 }
 
 // Value implements the driver.Valuer interface for database serialization.
-func (d *Decimal) Value() (driver.Value, error) {
-	if d == nil {
-		return nil, nil
-	}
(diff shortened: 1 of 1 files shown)


james-maloney:
Is this project still being maintained?

victorquinn:
Hey @james-maloney apologies for the long turnaround time here. I recently took over as maintainer of this repo so issue should be resolved much more quickly going forward.

I basically re-implemented this in 77e6b496f6c093115cb88318c1666a304989793a without changing the interface for `Value()` like you did here. Also, added a unit test for float32 to ensure it works as expected. 

So I'm closing this PR, but only because I manually put your changes into master. So thanks for the contribution, this was very helpful! (that just seemed faster/easier than bugging you to make the requested changes)

james-maloney:
Thanks!

PR #42, opened 2017-01-16
Expose a Coefficient() function

chrisguiney:
#20 did this as well, but seemed abandoned by the author.  Pretty essential when working with other decimal systems.


victorquinn:
Looks great, thanks!

PR #44, opened 2017-02-01
Added method Equal instead of method Equals

bgaifullin:
Motivation: the reflect.DeepEqual does not work for decimal, since the internal representation 
can be different for 2 same values.
I also make patch for testify/asserts to use method 'Equal' if structs implements this method to test
equality of objects. I choose name 'Equal' because the standard type time.Time has this method.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -385,9 +385,14 @@ func (d Decimal) Cmp(d2 Decimal) int {
 	return rd.value.Cmp(rd2.value)
 }
 
-// Equals returns whether the numbers represented by d and d2 are equal.
-func (d Decimal) Equals(d2 Decimal) bool {
+// Equal returns whether the numbers represented by d and d2 are equal.
+func (d Decimal) Equal(d2 Decimal) bool {
 	return d.Cmp(d2) == 0
+}
+
+// Equals is deprecated, please use Equal method instead
+func (d Decimal) Equals(d2 Decimal) bool {
+	return d.Equal(d2)
 }
 
 // Exponent returns the exponent, or scale component of the decimal.
--- a/decimal_test.go
+++ b/decimal_test.go
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"encoding/xml"
 	"math"
-        "math/big"
+	"math/big"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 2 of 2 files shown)


bgaifullin:
@victorquinn could you please review this PR, I need this patch to start using type Decimal in my project.

decimal.go
```diff
@@ -385,7 +385,12 @@ func (d Decimal) Cmp(d2 Decimal) int {
...
 
-// Equals returns whether the numbers represented by d and d2 are equal.
+// Equal returns whether the numbers represented by d and d2 are equal.
+func (d Decimal) Equal(d2 Decimal) bool {
+	return d.Cmp(d2) == 0
+}
+
+// Equals is deprecated, please use Equal method instead
 func (d Decimal) Equals(d2 Decimal) bool {
 	return d.Cmp(d2) == 0
```
victorquinn: Please update `Equals` to call `Equal` internally rather than duplicating the code

bgaifullin: @victorquinn Done. 

I changed Equals to Equal, because I also implemented code in asserts library to detect that structure implements method Equal and to use this method to compare objects instead of calling reflect.DeepEqual. 
there is my PR to asserts library https://github.com/stretchr/testify/pull/394

bgaifullin:
I updated PR

.idea/libraries/GOPATH__decimal_.xml
```diff
@@ -0,0 +1,97 @@
...
+      <root url="file:///usr/local/go/src/container" />
+      <root url="file:///usr/local/go/src/mime" />
+      <root url="file:///usr/local/go/src/path" />
+      <root url="file:///usr/local/go/src/reflect" />
+    </SOURCES>
+    <excluded>
+      <root url="file://$PROJECT_DIR$" />
+    </excluded>
+  </library>
+</component>
```
victorquinn: Thank you for the update, but this file should not have been checked in

bgaifullin: I am very sorry, the my code editor added this file automatically. :(

bgaifullin:
I have updated commit.

victorquinn:
Thanks @bgaifullin !

bgaifullin:
Thanks @victorquinn

PR #32, opened 2016-11-02
Added NullDecimal. Changed Decimal to implement driver.Valuer.

craigjackson:
From comments in PR 16.

PR #45, opened 2017-02-06
Added code to scan null values from the database

knifeedge:
Ran into the case where I had null values in a database and when it tried to scan I would get errors trying to scan a null type. Fixed this by returning a Decimal that is initialized with a 0 value and 0 exponent.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -579,7 +579,14 @@ func (d Decimal) MarshalJSON() ([]byte, error) {
 
 // Scan implements the sql.Scanner interface for database deserialization.
 func (d *Decimal) Scan(value interface{}) error {
-	// first try to see if the data is stored in database as a Numeric datatype
+	// check to see if we have a nil value, return a zero value and exp
+	if value == nil {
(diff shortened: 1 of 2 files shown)


victorquinn:
Hey @knifeedge I think this should be solved with #32 . Mind checking it out and updating here if it's not?

knifeedge:
Sounds good @victorquinn I'll look at this tonight.

knifeedge:
@victorquinn Yes I think that #32 will solve this problem, though it seems more work to have to deal with a NullDecimal type and check it's Valid field, versus just getting back a zero Decimal. If everyone is good with it, then let's go ahead with #32.

victorquinn:
@knifeedge yes, I generally agree that it's a bit easier to just get back a zero Decimal, but the change in #32 brings Decimal back into conformance with the `database/sql` `driver.Valuer` interface which I think is a better option overall.

knifeedge:
@victorquinn Agreed, I'll close this, thanks!

PR #48, opened 2017-02-17
Add Neg method

edsrzf:
Neg provides an easy way to negate a Decimal.


victorquinn:
Looks great, thanks for the addition!

PR #52, opened 2017-05-17
Defensive Copy on Coefficient

mattdee123:
Currently, if someone gets the Coefficient, and then modifies the value in the pointer (which is easy to do with the weird way the `big` package works), then it will change the underlying decimal, which can be a very hard bug to track down.  This fixes that.


victorquinn:
Great catch @mattdee123 !

PR #56, opened 2017-07-28
Added NewFromBigInt constructor & unit test

brobits:
I use this constructor for converting to/from this decimal type using raw values.

Example:

```go
d := decimal.New...
// extract raw values from decimal
signScale := d.Exponent() * int32(d.Sign()) // int32
bytes := d.Coefficient().Bytes() // []byte
```

```go
// create decimal from raw values
i := big.Int{}
i.SetBytes(b)
d := decimal.NewFromBigInt(&i, signScale)
```

Using `New(value int64, exp int32)` results in a loss of precision when the decimal's coefficient is larger than 8 bytes.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -81,6 +81,14 @@ func New(value int64, exp int32) Decimal {
 	}
 }
 
+// NewFromBigInt returns a new Decimal from a big.Int, value * 10 ^ exp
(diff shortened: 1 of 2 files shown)


victorquinn:
Looks great, thanks for the enhancement @brobits !

brobits:
thanks!

Issue #57, opened 2017-08-15
UnmarshalJSON throws error for null


klokare:
When JSON data contains a null value for a Decimal field, the UnmarshalJSON function returns an error: Error decoding string 'null': can't convert null to decimal.

I am unsure if this is the intended behaviour. For Go native type float64, a null is set to 0.0 when unmarshaling. In this [issue](https://github.com/golang/go/issues/9037) related to time.Time, the solution is to unmarshal null into the zero value. 

To me, it makes sense to unmarshal null to the zero value of Decimal when the target variable is a struct. If the user wants to unmarshal to nil, she can still make the target variable a pointer:

```Go

type Example struct {
    AsStruct   decimal.Decimal
    AsPointer *decimal.Decimal
}

var data = `{"AsStruct": null, "AsPointer"; null}`
var e Example

_ = json.Unmarshal(&e) // Ignoring error for this example. Actually, there would be no error returned.

fmt.Println("AsStruct", e.AsStruct, "AsPointer", e.AsPointer)
```

would produce 

```
AsStruct 0 AsPointer nil
```

victorquinn:
Fixed by #58

PR #58, opened 2017-08-15
Add null support to UnmarshalJSON

klokare:
Unmarshal null into zero value. Address Issue #57. Modifies UnmarshalJSON method and includes test.


victorquinn:
Looks great, thanks @klokare !

PR #64, opened 2017-10-18
#68: Fix Floor() and Ceil() for integer values

tbrs:
In some cases their results were different for otherwise equal Decimal instances. Bug was triggered by setting exponent to a positive value. For example, `New(1, 1).Floor()` returned 1 (i.e. `New(1, 0)`) instead of 10.

### DIFF
--- a/decimal_test.go
+++ b/decimal_test.go
@@ -508,11 +508,17 @@ func TestDecimal_rescale(t *testing.T) {
 }
 
 func TestDecimal_Floor(t *testing.T) {
-	type testData struct {
+	assertFloor := func(input, expected Decimal) {
+		got := input.Floor()
+		if !got.Equal(expected) {
+			t.Errorf("Floor(%s): got %s, expected %s", input, got, expected)
+		}
+	}
+	type testDataString struct {
 		input    string
 		expected string
 	}
-	tests := []testData{
+	testsWithStrings := []testDataString{
 		{"1.999", "1"},
 		{"1", "1"},
 		{"1.01", "1"},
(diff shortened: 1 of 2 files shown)


decimal_test.go
```diff
@@ -508,11 +508,17 @@ func TestDecimal_rescale(t *testing.T) {
 }
 
 func TestDecimal_Floor(t *testing.T) {
-	type testData struct {
+	assert := func(input, expected Decimal) {
```
victorquinn: I would prefer this be renamed to something a bit more descriptive like `assertFloor` as it's a bit difficult to tell what it's being used for further down in this doc. 

Further, `assert` tends to mean a specific thing in unit tests and calling this specific function `assert` when it's not a more generic assert but rather a specific one for this test means there's a higher likelihood of confusion.

decimal_test.go
```diff
@@ -525,22 +531,66 @@ func TestDecimal_Floor(t *testing.T) {
...
+	input := New(200, -2)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		input.Floor()
 	}
 }
 
 func TestDecimal_Ceil(t *testing.T) {
-	type testData struct {
+	assert := func(input, expected Decimal) {
```
victorquinn: Same comment as above, maybe call this `assertCeil`?

decimal_test.go
tbrs: Completely agree on this, thanks for extra details about the reasoning.

decimal_test.go
tbrs: Sounds good to me, will rename.

tbrs:
Thank you for reviewing! Naming updated as suggested.

PR #70, opened 2017-11-27
format usage code in README

ceshihao:
Usage codes in README are not well formatted.

Update it using gofmt.


victorquinn:
Thanks for the addition @ceshihao !

Issue #11, opened 2015-09-02
NewFromInt, NewFromInt32 and other convenience constructors


pascallouisperez:
Have you considered having simpler constructors

```
NewFromInt(v) = New(int64(v), 0)
```

etc.?

chakrit:
Or just `decimal.FromInt()` or `decimal.FromFloat()` would've suffice.

vadimg:
I have nothing against having more NewFromX constructors, and I welcome pull requests!

PR #72, opened 2017-12-13
Add NewFromInt

mathieupost:
Resolves https://github.com/shopspring/decimal/issues/11

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ func main() {
 		panic(err)
 	}
 
-	quantity := decimal.NewFromFloat(3)
+	quantity := decimal.NewFromInt(3)
 
 	fee, _ := decimal.NewFromString(".035")
 	taxRate, _ := decimal.NewFromString(".08875")
--- a/decimal.go
+++ b/decimal.go
@@ -87,6 +87,32 @@ func New(value int64, exp int32) Decimal {
 	}
 }
 
+// NewFromInt converts a int64 to Decimal.
+//
+// Example:
+//
+//     NewFromInt(123).String() // output: "123"
+//     NewFromInt(-10).String() // output: "-10"
+func NewFromInt(value int64) Decimal {
+	return Decimal{
+		value: big.NewInt(value),
+		exp:   0,
+	}
+}
+
+// NewFromInt32 converts a int32 to Decimal.
+//
+// Example:
+//
+//     NewFromInt(123).String() // output: "123"
+//     NewFromInt(-10).String() // output: "-10"
+func NewFromInt32(value int32) Decimal {
+	return Decimal{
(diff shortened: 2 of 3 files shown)


victorquinn:
First, thanks for the additions!

However, is there a reason you chose `newFromInt` which takes an `int64` and `newFromInt32` which takes an `int`?

Why not `newFromInt` which takes an `int` and `newFromInt64` which takes an `int64`?

If there are precedents elsewhere for this kind of thing elsewhere I could be convinced that it's a good idea, but on my first glance it seems rather confusing to have `newFromInt` which doesn't take an `int` and `newFromInt32` which takes an `int` and no `newFromInt64`

victorquinn:
Just following up here, any reply to the above @mathieupost ?

mathieupost:
Sorry for the late reply, I'm on a holiday at the moment ;)
I named it this way because that was how it was named in https://github.com/shopspring/decimal/issues/11.
And I thought that since int64 is used more as far as I know, it would be cleaner in most code to be able to simply use newFromInt for those.

mathieupost:
@victorquinn any comment on this?

Mungrel:
I agree that `NewFromInt(v int)`, `NewFromInt32(v int32)` and `NewFromInt64(v int64)` would be a more intuitive API.

mathieupost:
Since NewFromFloat uses float64 and most libraries use int64 for integers I'll just keep the NewFromInt(int64) method and remove the NewFromInt32(int32) method, since then it'll be consistent with the rest of the API of this library.

Mungrel:
Looks like you included your `.idea/` directory

mathieupost:
Oops, fixed ;)

k1ng440:
please merge it

mwoss:
Hi @mathieupost, a new maintainer here. Thanks for your contribution. 
May I know what will be the purpose of `NewFromInt` method when `New` method exactly matches the current use case?
If we want to create a more flexible API, maybe we should add `NewFromInt32` too? What do you think?
FYI. @njason

mathieupost:
Hi! This is a long time ago 😅 

I guess the idea was that with the `New` method you have to specify the exponent to `0` which is then kind of a magic number in the code of people using this lib. To me `NewFromInt` would be more clear when reading code where this lib is used.

mwoss:
I think your point is pretty valid. 
What's your opinion about adding both `NewFromInt32` and `NewFromInt` initializer methods? It will much already existing FromFloat methods.

mathieupost:
I could add `NewFromInt32` again. I guess it would indeed be nice to match the Float methods.

mwoss:
IMO, looks legit. : ) Let's wait for Jason's feedback

PR #73, opened 2017-12-30
RequireFromString

randyp:
**Proposal**
decimal.RequireFromString which delegates to decimal.NewFromString and panics if an error was returned. Please rename the function to your liking, possibly to "MustParseFromString".

**Reasoning**
We often use decimal.NewFromString with string constants that are guaranteed to not cause errors. We should have the power to ignore the error scenario for string constants. The regexp package exposes [MustCompile](https://golang.org/pkg/regexp/#MustCompile) for the same reason.

**Why Not "Must"**
Another pattern is the "Must" pattern, such as [template.Must](https://golang.org/pkg/html/template/#Must). Not my preference, but one or the other could be in this library. Three lines of code here will save three per user of your library.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -153,6 +153,22 @@ func NewFromString(value string) (Decimal, error) {
 	}, nil
 }
 
+// RequireFromString returns a new Decimal from a string representation
+// or panics if NewFromString would have returned an error.
+//
+// Example:
+//
+//     d := RequireFromString("-123.45")
+//     d2 := RequireFromString(".0001")
+//
+func RequireFromString(value string) Decimal {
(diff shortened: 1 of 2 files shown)


jterry75:
+1 - I intuitively always want to use this in a struct initializer like:
```go
type Test struct {
   Value Decimal
}

func main() {
    t := Test{
        Value: decimal.NewFromString("123.23")
    }
}
```

But because of the error return I have to pull them out and assign them after creation. I am all for the panic because I have already validated that the inputs are correct so the data in the string is trusted.

victorquinn:
Woohoo great addition @randyp apologies it took so long to resolve!

Issue #75, opened 2018-01-18
RoundBank produces incorrect results depending on Decimal's internal representation


freakingawesome:
The `RoundBank()` function, given the decimal value `2.5`, should return `2`. This works as expected when your decimal is represented by `New(25, -1)`.

However, when represented by `New(250, -2)`, the function `RoundBank()` gives the incorrect result of `3`.

Here is a currently failing test which illustrates the problem:

```golang
func TestRoundBankAnomaly(t *testing.T) {
    a := New(25, -1)
    b := New(250, -2)

    if !a.Equal(b) {
        t.Errorf("Expected %s to equal %s", a, b)
    }

    expected := New(2, 0)

    aRounded := a.RoundBank(0)
    if !aRounded.Equal(expected) {
        t.Errorf("Expected bank rounding %s to equal %s, but it was %s", a, expected, aRounded)
    }

    bRounded := b.RoundBank(0)
    if !bRounded.Equal(expected) {
        t.Errorf("Expected bank rounding %s to equal %s, but it was %s", b, expected, bRounded)
    }
}
```

freakingawesome:
I stumbled across this bug when using `RoundBank` on the result of decimal multiplication (which may increase the `exp` of `Decimal`), which leads me to believe that the incorrect results from this bug are subtle and widespread.

orian:
@freakingawesome do you mean the bug is still there? Could you give some extra examples?

freakingawesome:
No, the bug is now fixed. What I meant is that it was a subtle bug that has existed for years, so loads of calculations throughout the years could be slightly incorrect, especially is you were to run comparisons against the same set of input, before and after the bug was fixed.

orian:
Thanks for clarification.

mwoss:
@freakingawesome Thanks for explanations of this bug and fixing it :)) I'm closing the issue.

PR #76, opened 2018-01-18
Fix broken RoundBank func in #75

freakingawesome:
This fixes the issue described in #75, wherein `RoundBank()` can return incorrect results for the same decimal value, varying based on the internal representation.


victorquinn:
Awesome fix, you've lived up to your username thanks @freakingawesome

PR #80, opened 2018-01-26
Added ensureInitialized() to the Neg() method

ivahaev:
When d.value is nil `Neg()` will panic. This change fixed it.


victorquinn:
Looks great, thanks for the fix @ivahaev !

PR #86, opened 2018-03-16
Add support for shifting in base 10

stevenroose:
This allows for very fast multiplying and dividing by powers of 10
without the possibility of losing precision.

We've personally experienced weird rounding values when dividing by 100.


stevenroose:
I assumed checking the int32 for overflows was not necessary.

victorquinn:
Nice addition @stevenroose , thanks!

PR #88, opened 2018-03-22
add functions IsPositive, IsNegative and IsZero

jdaarevalo:
When we work with numbers, it is often necessary to know if these are positive, negative or zero.
Although the Sign function is available, I consider it valuable to have these new functions that help validate if a decimal is Positive, Negative or Zero.
regards


victorquinn:
Excellent addition, apologies it took so long to merge!

PR #96, opened 2018-06-07
Added cos, sin, tan, and atan methods and tests

i-norden:
I needed to do some trig with decimals so I added cos, sin, tan, and atan methods and their tests.


i-norden:
Odd, go test -v exits with all passes and no errors on my end. 

The error thrown in the CI build/test is:
./decimal_test.go:261:6: d declared but not used

That line shouldn't have been affected by my additions.

i-norden:
Any interest in these additions?

peterstace:
(not the maintainer, but use the decimal lib quite a lot)

Since the constants used for the trig tables are only in double precision, is that limiting the accuracy of the output? What sort of accuracy gain do we get with these changes compared to just converting a decimal back to a float64, running the regular trig functions on them, then converting back to decimal?

i-norden:
The results from the calculations performed using those double bit precision constants will have decimal precision. If any of the internal calculations produce a result outside of double bit precision it can now retain that precision and propagate it throughout the rest of the calculations and into the final output, resulting in overall greater precision and accuracy. 

At least, that's how I understand it. I can also try to find greater precision constants to start with and load them in from strings instead.

victorquinn:
Awesome addition, thanks @i-norden !

Issue #97, opened 2018-06-12
Got wrong result Round()


yusank:
``` go
func Decimal() {
    f, exit := decimal.NewFromFloat(1.275).Round(2).Float64()
	if exit {
		fmt.Println("yes")
	}

	fmt.Println(f)
        // result should be 1.28, but got 1.27
}
```

I'm still new to Go, so why I got what I don't want, can anybody explain ,thx.

Is there any way to fix this?

klauspost:
Should be fixed by #100

leighmcculloch:
This does appear to be fixed in the latest commit. e.g. https://play.golang.org/p/3y7fw2KjHjk

mwoss:
Apparently it was fixed. Closing the issue.

PR #100, opened 2018-07-09
Make NewFromFloat respect the precision of the input

klauspost:
Restores the previous behaviour where input converted from float is truncated at the precision of the float.

The precision is depending on the actual value. Simply making it 15 digits would make it faster, but would in some cases lose some precision. So the code from the stdlib that does this calculation (very well) has been included.

Lots of good articles here: https://www.exploringbinary.com/decimal-precision-of-binary-floating-point-numbers/

Performance is around the same as the previous string roundtrip since it basically does the same, but allocations are a bit less.

`BenchmarkNewFromStringFloat` is the old method, `BenchmarkNewFromFloat` is the new.

```
BenchmarkNewFromFloatWithExponent-8     10000000               260 ns/op             174 B/op          4 allocs/op
BenchmarkNewFromFloat-8                  2000000               744 ns/op              90 B/op          2 allocs/op
BenchmarkNewFromStringFloat-8            2000000               822 ns/op             258 B/op          6 allocs/op
```


klauspost:
Rebased+squashed.

I don't know if the Go License itself should be included. I can add a `GO_LICENCE` file and reference that from the go code, but I am no license expert.

I see the new tests rely on the new values. I will adjust them.

This should also fix unpredictability like #97

victorquinn:
FYI this is available and tagged as `1.1.0`

PR #110, opened 2018-08-19
Create numbers.go

kempeng:
Numbers.go defines several values in decimal, extending on the Zero var already defined. These var's make the use of numbers value in decimal representation easier to use and read

### DIFF
--- a/numbers.go
+++ b/numbers.go
@@ -1,0 +1,55 @@
+package decimal
+
+// MinusOne defines value -1 in decimal format
+var MinusOne = New(-1, 0)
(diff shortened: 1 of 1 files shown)


kempeng:
can you explain why this testing failed?

igrmk:
@kempeng, it looks like you defined `Ten` two times. One should be `Twenty`.

kempeng:
@igrmk , yes you're right, it should be Twenty, apologies

kempeng:
fixed in new request

PR #115, opened 2018-08-29
Running gofmt

igrmk:
Somehow gofmt was not run for some code.
It is hard to work with it.
For example my vim do it automatically and it leads to huge diffs.


mwoss:
I have already run `gofmt` some time ago on the entire project, so I will close this PR, always thanks a lot for your contribution :3.

PR #120, opened 2018-10-06
RoundStep method

wildSeagull:
While working with a few currency exchanges' APIs I noticed that some of them specify the exchange price incrementation number (aka step), so in this PR I've added a method that rounds the decimal to the nearest step value.


dez11de:
Ran into the same issue today, and ghosts solution looks good to me. Is there something I can help with in getting this merged?

PR #121, opened 2018-10-06
IfThen and NoZero methods

wildSeagull:
IfThen and NoZero are helper methods to be used when in-line checks and value changes need to be performed e.g. number division by a calculation that might result in zero.


mwoss:
Personally, I think those helper methods are unintuitive and unnecessary for the decimal library. 
Such things should not be a part of the core library. Closing.

PR #124, opened 2018-10-14
Added support for Decimal places in JSON marshall

IvRRimum:
This seems super usefull.

Basically allows user to set how many numbers after decimal to return(and rounds up).
Using `StringFixed`.

Thanks!


adevjoe:
It is usefull for me. I have the same implementation.No one accepts it?

theoribeiro:
Sorry to bring up a zombie 6 year old PR back but any chance something like this might be accepted @mwoss?

PR #127, opened 2018-12-14
go fmt

cryptix:
sorry to waltz over the history. Is there a reason that file isn't gofmted?


njason:
Taken care of in #162

PR #139, opened 2019-03-03
Update decimal.go

lanre-ade:


decimal.go
```diff
@@ -970,6 +970,11 @@ func (d Decimal) MarshalBinary() (data []byte, err error) {
 
 // Scan implements the sql.Scanner interface for database deserialization.
 func (d *Decimal) Scan(value interface{}) error {
+	if value == nil {
+		d = nil
```
serprex, line 974:
```suggestion
		*d = nil
```

serprex: Except this won't work because `*d` expects a Decimal

PR #141, opened 2019-04-06
decimal: add decomposer interface

kardianos:
For golang/go#30870

### DIFF
--- a/decomposer.go
+++ b/decomposer.go
@@ -1,0 +1,47 @@
+package decimal
+
+import (
+	"fmt"
+	"math/big"
+)
+
+// Decompose returns the internal decimal state into parts.
+// If the provided buf has sufficient capacity, buf may be returned as the coefficient with
+// the value set and length set as appropriate.
+func (d Decimal) Decompose(buf []byte) (form byte, negative bool, coefficient []byte, exponent int32) {
+	negative = d.value.Sign() < 0
+	exponent = d.exp
+	coefficient = d.value.Bytes()
+	return
+}
+
+const (
+	decomposeFinite   = 0
+	decomposeInfinite = 1
+	decomposeNaN      = 2
+)
+
+// Compose sets the internal decimal value from parts. If the value cannot be
+// represented then an error should be returned.
+func (d *Decimal) Compose(form byte, negative bool, coefficient []byte, exponent int32) error {
+	switch form {
+	default:
+		return fmt.Errorf("unknown form: %v", form)
+	case decomposeFinite:
+		// Set rest of finite form below.
+	case decomposeInfinite:
+		return fmt.Errorf("Infinite form not supported")
+	case decomposeNaN:
+		return fmt.Errorf("NaN form not supported")
+	}
+	// Finite form.
+	if d.value == nil {
+		d.value = &big.Int{}
+	}
+	d.value.SetBytes(coefficient)
+	if negative && d.value.Sign() >= 0 {
+		d.value.Neg(d.value)
+	}
+	d.exp = exponent
+	return nil
+}
--- a/decomposer_test.go
+++ b/decomposer_test.go
@@ -1,0 +1,90 @@
+package decimal
+
+import (
+	"testing"
+)
+
+func TestDecomposerRoundTrip(t *testing.T) {
+	list := []struct {
+		N string // Name.
+		S string // String value.
+		E bool   // Expect an error.
+	}{
+		{N: "Normal-1", S: "123.456"},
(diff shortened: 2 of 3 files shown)


kardianos:
Please let me know what you think of this idea. See referenced issue.

decomposer.go
```diff
@@ -0,0 +1,84 @@
...
+		i++
+		j--
+	}
+	return
+}
+
+// Compose sets the internal decimal value from parts. If the value cannot be
+// represented then an error should be returned.
+func (d *Decimal) Compose(form byte, negative bool, coefficient []byte, exponent int32) error {
+	switch form {
```
njason: These forms should constants rather than magic numbers

decomposer.go
```diff
@@ -0,0 +1,84 @@
...
+	default:
+		return fmt.Errorf("unknown form: %v", form)
+	case 0:
+		// Set rest of finite form below.
+	case 1:
+		return fmt.Errorf("Infinite form not supported")
```
njason: Why even include these cases, or have the form argument at all, if they cannot be accepted?

decomposer_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"testing"
+)
+
+func TestDecomposerRoundTrip(t *testing.T) {
+	list := []struct {
+		N string // Name.
+		S string // String value.
+		E bool   // Expect an error.
+	}{
+		{N: "Normal-1", S: "123.456"},
```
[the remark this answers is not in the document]
kardianos: I don't see any uses of an "assert" package, but, I added more test cases, it already checks with Decimal.Cmp and I added a check to the original S. Because this is a round trip test, I'm checking the final decimal against the original parsed number.

kardianos:
I removed the use of testing.T.Run, which was introduced in Go1.7. CI passes now.

kardianos:
Ping.

kardianos:
Ping

kardianos:
@njason Can we resume this conversation? This is in Go1.13 and crdb/apd and ericlagergren/decimal.

njason:
@kardianos I see this has been added as an experimental interface to 1.13 which is in an unstable release state. Please circle back when this is in a stable release.

kardianos:
1.13 has now been released.

njason:
Can we remove the `decomposer` interface in this PR since it's already defined in [core](https://github.com/golang/go/blob/5ff38e476177ce9e67375bd010bea2e030f2fe19/src/database/sql/convert.go#L583)?

kardianos:
The most recent build on TravisCI complained about Go1.2 being missing. Looking on https://golang.org/dl/ Go 1.2.2 is present, but Go1.2 is indeed gone. To test, I changed the travis ci file and it passed using Go1.2.2.

kardianos:
Also, I removed the interface per your request.

5teven:
Hello everyone

After this update i got error:
`pq: encode: unknown type for decimal.Decimal`

I'm using:
go1.13 linux/amd64
postgresql 10
sqlx

kardianos:
@5teven What is the postgresql driver version? Also, please open a new issue at golang.org/issue and @ me.

tietang:
After this update i got error:
`cannot convert type: decimal.Decimal`

go 1.13.4
go-sql-driver/mysql 1.4.1

i think go mysql driver can't  support decimalDecompose.

please see:
1. database/sql/driver.IsValue, line 176:
`
// IsValue reports whether v is a valid Value parameter type.
func IsValue(v interface{}) bool {
	if v == nil {
		return true
	}
	switch v.(type) {
	case []byte, bool, float64, int64, string, time.Time:
		return true
	case decimalDecompose:
		return true
	}
	return false
}
`

2.go-sql-driver/mysql  statement.go line 140, method ConvertValue:
line 141:
`
if driver.IsValue(v) {
		return v, nil
	}

	if vr, ok := v.(driver.Valuer); ok {
		sv, err := callValuerValue(vr)
		if err != nil {
			return nil, err
		}
		if !driver.IsValue(sv) {
			return nil, fmt.Errorf("non-Value type %T returned from Value", sv)
		}
		return sv, nil
	}
`

3. go-sql-driver/mysql  packets.go line 1071

line 1071
`default:
				return fmt.Errorf("cannot convert type: %T", arg)
			}`

njason:
commit 75bb2cb has been reverted from master and will be worked on in another branch

tsingsun:
any updated?  decimalDecompose interface has been in go 1.17,whether support can be considered

PR #144, opened 2019-06-17
fix: Coefficient panic

forsaken628:
call Coefficient on uninitialized Decimal will panic


bmperrea:
@njason We've been working around this for a bit in our projects by adding early outs if the decimal is zero before calling Coefficient. Please take a look! A simple test is `decimal.Zero.Coefficient() // panics`

forsaken628:
Temporary solution

d = d.Shift(0)
bi := d.Coefficient()

PR #66, opened 2017-10-19
Implement Swedish/Cash rounding fixes #63

SchumacherFM:
Hey,

I need the following guidance: Have I implemented the rounding correctly? Are there any architecture or other bugs? Is the panic correct? Are comments correct and consistent? How can I (or you) reduce the allocations? There seems to be some tricks in the library to avoid allocations but I'm not familiar with the math/big package. You can also work on the allocs to reduce them. Let me know.

RoundCash aka Cash/Penny/öre rounding rounds decimal to a specific
interval. The amount payable for a cash transaction is rounded to the nearest
multiple of the minimum currency unit available. The following intervals are
available: 5, 10, 15, 25, 50 and 100; any other number throws a panic.
```
    5:   5 cent rounding 3.43 => 3.45
   10:  10 cent rounding 3.45 => 3.50 (5 gets rounded up)
   15:  10 cent rounding 3.45 => 3.40 (5 gets rounded down)
   25:  25 cent rounding 3.41 => 3.50
   50:  50 cent rounding 3.75 => 4.00
  100: 100 cent rounding 3.50 => 4.00
```
For more details: https://en.wikipedia.org/wiki/Cash_rounding

```
BenchmarkDecimal_RoundCash/five-4         	 1000000	      1918 ns/op	    1164 B/op	      30 allocs/op
BenchmarkDecimal_RoundCash/fifteen-4      	  300000	      4331 ns/op	    2940 B/op	      74 allocs/op
```

Issue #107, opened 2018-08-03
Zeros trimmed off when parsing strings, leading to inability to use string to store a decimal value


imirkin:
Looks like merge #46 (Remove insignificant digits during string parsing) made it impossible to initialize decimals from string with a fixed precision. What was the motivation for that change, and would a change be accepted to undo it, or move it off into a NewFromStringTrimmed() function? [Or, less preferably, introduce a NewFromStringPrecise variant.]

The whole point of decimals is to be precise, and there is a semantic difference between 1.0 and 1.00.

imirkin:
Oh wow! It looks like .String() was also changed to cut off zeroes. That goes counter to how any other bigdecimal library works.

Do let me know if this is something that you'd be interested in having fixed, or if I need to clone this off.

wizardist:
@imirkin could you tell how exactly does this make it impossible? Examples and/or repro code would be highly appreciated.

wizardist:
So far looks like a presentation issue, but I think I understand your concern. If you're not storing decimals as strings in DB, then you're safe.

For presentation purposes you have methods like `StringFixed()` and a few others which may or may not help you.

HTH

imirkin:
Nope, it's not a presentation issue. 1.0 and 1.000 are semantically different values and I need to be able to tell which is which. Feel free to close this issue, I'll just clone this off.

kempeng:
Do I understand your issue correctly in that you want the first value (1.0) to be stored as (1, 0) (bigInt value, precision) and the second one as (10000, -4) ? I think that is a valid ask to be supported by the package?

I stumbled on this in a related application of the decimal package in my code: I created new Decimals with values in multiples of a thousand (1000, 2000, 5000, 10000) by creating a new decimal's with the value as big int value and precision as zero). However later on in the code, I divided the values by a thousand, and only realised later that the precision was not zero anymore (as I naively expected), but had become -16, the default precision of DivRound() ....

imirkin:
A decimal is the composition of a big integer and an exponent, (a, b). It would be nice to be able to save those off to a string, and then load them again from the string, such that one gets the same values back of (a, b). Otherwise the values can't be serialized.

The globally agreed string representation of all this is the logical one... write out the big integer, and then insert a . in the right place [or potentially eX if one tries to represent e.g. (1, 2)].

[...]

Issue #89, opened 2018-03-28
RoundCash interval Is 15


candy4dmu:
`fmt.Println(decimal.NewFromFloat(3.5).RoundCash(15))`

the result Is 2.4

`fmt.Println(decimal.NewFromFloat(3.45).RoundCash(15))`

the result is 3.5 , not same with the doc.

mwoss:
@candy4dmu Thanks for pointing this issue. You are correct, something is not working correctly. We will try to fix it as soon as possible.

PR #159, opened 2019-12-10
Revert remove insignificant digits

njason:
When I went to revert #46, `TestDecimal_RoundCash` started failing because the `RoundCash` function  (#66) relies on a Decimal with insignificant digits removed. To get around this problem I made a new function to remove insignificant digits

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -153,13 +153,8 @@ func NewFromString(value string) (Decimal, error) {
 		// an int
 		intString = value
 	} else if len(parts) == 2 {
-		// strip the insignificant digits for more accurate comparisons.
-		decimalPart := strings.TrimRight(parts[1], "0")
-		intString = parts[0] + decimalPart
-		if intString == "" && parts[1] != "" {
-			intString = "0"
-		}
-		expInt := -len(decimalPart)
+		intString = parts[0] + parts[1]
+		expInt := -len(parts[1])
 		exp += int64(expInt)
 	} else {
 		return Decimal{}, fmt.Errorf("can't convert %s to decimal: too many .s", value)
@@ -852,6 +847,7 @@ func (d Decimal) RoundBank(places int32) Decimal {
 // 	  100: 100 cent rounding 3.50 => 4.00
 // For more details: https://en.wikipedia.org/wiki/Cash_rounding
 func (d Decimal) RoundCash(interval uint8) Decimal {
+	iD := d.RemoveInsignificantDigits()
 	var iVal *big.Int
 	switch interval {
 	case 5:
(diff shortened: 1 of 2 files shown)


mwoss:
Currently, there is a bug in `RoundCash` method, mentioned in #89.  We have to fix this issue before merging this PR.

philippgille:
@njason @mwoss 

Thanks for maintaining this library! A quick question regarding this PR:

Is it correct that when using `Decimal` in a struct and encoding an object of that struct and later decoding it (for example during persistence and retrieval from a DB) and comparing the two objects in a test with something like `require.Equal()` (which uses `reflection.DeepEqual()` under the hood), that after this PR the objects are seen as unequal, whereas before they were seen as equal?

And the correct way to compare them would be to use `String()` on both objects?

mwoss:
Hi @philippgille. I apologize for the late response, I had a tough weekend. 
The scenario you've described should not have happened. Are you using correct struct tags so values can be properly serialized/deserialized? Could you give us a small code snippet to reproduce this issue?

philippgille:
For creating a snippet I had another look and the root cause seems to be that the initial decimal is created with `Round(6)`, while the Postgres database column is `decimal(x, 8)`. So after fetching the number from the DB and deserializing it, a decimal with 8 decimal places is created. But the last two digits are `00`, because the original decimal just had 6 decimal places. Now *before* this PR these two trailing zeroes were *removed* and thus the `reflect.DeepEqual()` check yielded `true`, while *after* this PR the trailing zeroes are not removed and thus the check yields `false`.

So in the end I'd say this library behaves correctly and as expected (8 decimal places in the DB lead to 8 decimal places in the deserialized field). It's just a breaking or at least noteworthy change.

mwoss:
@philippgille Yes, if those decimals differ by two last zeroes after a comma, they for sure, should not be equal. This was discussed in depth in #107.

philippgille:
Does it make sense to add a note to the README or so, because the behavior was different for almost exactly 3 years, so I can imagine other people updating the package and wondering about failing unit tests as well :)

mwoss:
I'm 100% agreeing with you. I'm adding this task to our backlog.
Thanks a lot :))

daemonfire300:
> I'm 100% agreeing with you. I'm adding this task to our backlog. Thanks a lot :))

I could not find the backlog issue so pardon me for reviving this zombie thread 😄 

https://github.com/shopspring/decimal/pull/277

tzachshabtay:
We just got burned by this breaking change after upgrading the package.
We need to know the precision ignoring trailing zeroes (i.e what `Exponent` used to do), my current workaround is:
```go
func getOldExponent(x decimal.Decimal) int32 {
	return decimal.RequireFromString(x.String()).Exponent()
}

```

Is there a better way? @mwoss  (p.s thank you for this library).

mwoss:
I deeply apologize for my late reply! I was not actively maintaining this library in the past year :<

---

@daemonfire300 - Sure I will take a look at your PR 

@tzachshabtay - I apologize for that. We should've communicated this change more broadly or introduced it in the 2.x release. Right now library does not provide an easy way to calculate precision ignoring trailing zeroes. You can do it your way or count the trailing zeroes and substract it from `Exponent`:
```go
func getPrecisionWithoutTrailingZeroes(d Decimal) int32 {
	coeff := d.Coefficient().String()
	trailingZeroes := len(coeff) - len(strings.TrimRight(coeff, "0"))
	return min(0, d.Exponent()+int32(trailingZeroes)) // for positive exponents decimal precision is always 0
}
```

PR #166, opened 2020-02-16
remove 15 interval RoundCash

njason:
Addresses: https://github.com/shopspring/decimal/issues/89

Original PR: https://github.com/shopspring/decimal/pull/66

Not sure why 15 interval was included in the original PR as there is no information online about it.  @SchumacherFM could you explain?


SchumacherFM:
@njason This is mentioned in the description of the pull request. To recite it:

```
10:  10 cent rounding 3.45 => 3.50 (5 gets rounded up)
15:  10 cent rounding 3.45 => 3.40 (5 gets rounded down)
```

Why do you want to remove it?

mwoss:
Firstly, RoundCash code returns incorrect values for interval = 15. We could dig in and fix it, but there is no point in doing that. I don't see any use case for such an interval. I've looked through dozens of wiki pages and other websites and I couldn't find any info about it.
Secondly, there was no real explanation of including 15 rounding interval at all. 
`15:  10 cent rounding 3.45 => 3.40 (5 gets rounded down)` is not an explanation, imo.  
From your example it looks like interval.= 15 should be used when we want round decimal to 10s, but the rounding result needs to be rounded down. 
Rounding down/up logic could be useful, but it should be implemented as an additional parameter to the method signature, not as interval.

SchumacherFM:
Thank you for the explanation. Approved ;-)

The PR should not only remove the code but also the comments indicating the 15 interval.

PR #171, opened 2020-04-19
Add new helper methods: BigInt, BigFloat

mwoss:
Those are requested by the community two new helper methods that would make it easier to use the library and prevent users from making an improper decimal casting. 
In the tests, I used strings for compersion as it is easier to read them in that way and it does not change result of the tests.


njason:
Or loss of precision

mwoss:
Hm, I wasn't aware that `big.Float` is not available in Go v.1.2. What should we do? Drop support for that version or remove that new functionality? I'm not sure whether `decimal` should be still compatible with over 7 years old version fo Go.

mwoss:
@njason `big.Float` is available from Go version >1.5. I would say it's ok to remove support for such old Go versions and stick to the pattern of other open-source go libraries

njason:
Sounds good to me

PR #172, opened 2020-04-19
Drop support for Go versions older than 1.7

mwoss:
As we want to introduce new helper methods that rely on functionalities available from Go version >1.5 we decided to drop support for old Go versions. 
We would still support as a bunch of older go version, but we have to think more strictly about moving this library forward.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.6.x
+  - 1.7.x
   - 1.12.x
   - 1.13.x
   - tip
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.5.x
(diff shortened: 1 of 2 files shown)


njason:
Should also update readme with go version reqs. I don’t think go.mod allows you to add supported go versions

mwoss:
Wow that's odd `TestNewFromFloatWithExponent` is failing on 1.5.4

mwoss:
Oh, there was a bug in `big` built-in lib in 1.5 version. I will try to check it on 1.6 via Travis as I had some problems with compiling this Go version locally. Lib is for sure working on Go 1.7 and 1.8, I checked it second ago.

mwoss:
Apparently, it looks like that bug was fixed in 1.7 :<

mwoss:
@njason could you check it? :D

PR #179, opened 2020-06-28
Implement IsInteger method

mwoss:
This pull request contains an implementation of the IsInteger method that I'm using in WIP implementation of the decimal power method. I'm introducing this functionality in separate PR as new Power implementation would be enough difficult to review and I don't want to hinders it anymore. 

I experimented a bit with performance here and I decided to go with the most straight forward implementation (simple abs function, simple checks) as the improvement was almost not visible at all, but they make implementation tricky to understand.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -583,6 +583,33 @@ func (d Decimal) Pow(d2 Decimal) Decimal {
 	return temp.Mul(temp).Div(d)
 }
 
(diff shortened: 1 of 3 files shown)


mwoss:
Yes, it's a pretty rare case. If I would see any performance benefits by introducing this case I would add it along with new Power implementation

Issue #145, opened 2019-06-17
NewFromString: Thousands Separator on Input


coip:
the string `$8,112.61` breaks for me. 

I am hesitant to open this given that the `$` is already *not* supported.

Should `FromString` variants handle string formatters? 
Maybe strip out `$`/`,` characters?

Hesitant to preemptively submit a PR given i18n implications, ie comma meaning fractional vs purely formatting in European countries.

Open to feedback, and submit PR for `s/[$,]//g` filtering if that's reasonable and aligned with the project goals.

Thanks!

coip:
Also, only found one other related issue, #98, which is related to this, though I believe that issue is for _formatting output_ whereas this is for _handling formatted input_

mwoss:
Hi @coip. Thanks for suggesting such functionality. 
Tbh, I'm not convinced of the necessity of handling money formatted string by `FromString` method.
Anyways, I'm aware that decimal libs are mostly used for money operations, so we will consider implementing this. (FYI. @njason)

coip:
> I'm not convinced of the necessity

I don’t blame ya. It is a bit out of the scope given the use-case specificity. Maybe considering it more broadly, as a `func FromFormattedString()` for example, would help it be more applicable for a ~parsing suite.

> we will consider implementing this.

Appreciate it, if I find some bandwidth I will try to open a feasible pr but that’s honestly not v likely at the moment.

mwoss:
NewFromFormattedString would be a better idea, IMO. We will add this task to backlog ;)

W-Floyd:
This has been resolved, `decimal.NewFromFormattedString` has existed for some time now. You can use it like so:
```
val, err := decimal.NewFromFormattedString(str, regexp.MustCompile("[$,]"))
```

PR #183, opened 2020-08-24
feat: adding in support for commas 

ConradKurth:
We have a situation where we are reading in values from a CSV using gocsv, we would like to have support for stripping out commas from numbers. We believe this is reasonable support since having numbers with commas is a reasonable annotation to numbers.

Would love to hear your thoughts or if my additional of a test is not satisfactory, let me know!

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -132,6 +132,7 @@ func NewFromString(value string) (Decimal, error) {
 	var intString string
 	var exp int64
 
+	value = strings.Replace(value, ",", "", -1)
 	// Check if number is using scientific notation
 	eIndex := strings.IndexAny(value, "Ee")
 	if eIndex != -1 {
--- a/decimal_test.go
+++ b/decimal_test.go
@@ -317,6 +317,7 @@ func TestNewFromStringDeepEquals(t *testing.T) {
 		{"10", "10.0", false},
 		{"1.1", "1.10", false},
 		{"1.001", "1.01", false},
+		{"12,000.00", "12000.00", true},
 	}
 
 	for _, cmp := range tests {


mwoss:
Hi @ConradKurth! Thanks for your contribution. Also, it's nice to hear that you are using `decimal` lib too.

I was thinking about your solution for a bit and I came to the conclusion that we should not merge it in the state like this.
The reasoning is straight forward. Merging this PR could be a cause of crashing dozens of pipelines that already use `decimal` library. Even though changes you introduced are fairly small, and at first glance should not be problematic for anyone, they are touching core initialization logic that probably every user of `decimal` lib is using. Devs that depended on how `NewFromString` worked could be surprised that method is not failing on values like `61,,,,,1` or even correct one like `5,199.2`.
By design minor version bump should not introduce any breaking changes - we know it's not always true, but let's not talk about it much :DD.
In my opinion, the best idea here would be to introduce an optional parameter for a `NewFromString` method, called `thousandSeparator`. It would keep backward compatibility and also give users the freedom to configure their own separator. Unfortunately, the Go language does not provide such capability, so it's a no-op, at least for now. 

So, I would propose is an introduction of `NewFromFormattedString` method (or something similar in naming) that would handle various decimal representation like those with thousand separators (12,000.00; 12_000.0), money signs ($5.12, €55.12), etc.
A similar idea was mentioned in PR #145. I kinda deprioritized it as I thought is not a necessary feature, but now I think more people need it than I expected. I will start the implementation this week. 
This should be released in version `1.3.0` after I finish unfamous `Pow` method reimplementation. 

In version `2.X` both `NewFromString` and `NewFromFormattedString` could be merged into one method. 

What do you think @ConradKurth?

mwoss:
The requested functionality has been implemented in #184

mwoss:
Due to no response and merged implementation #184 I'm closing this pull request.

Issue #98, opened 2018-06-28
Thousands separator


NathanZarro:
I don't see a format option for thousands separators, such as commas. Am I missing it?

peterstace:
No, you're not missing it, there is no such format option.

That would be useful for me as well (in my use case, displaying large numbers to users in a friendly way).

mwoss:
Thanks for pointing this @NathanZarro @peterstace. Let's move the conversation to #145. I'll close this issue.

PR #184, opened 2020-09-07
Implement new initialization function - NewFromFormattedString

mwoss:
#### Motivation: 
This new init function was requested more than 3 times (#98 #145 #183), so I thought it's the time we should implement such functionality. Also, personally I think it would be a great addition to the library too.

#### Implementation:
Implementation is pretty straight-forward. `NewFromFormattedString` is a wrapper around `NewFromString`, but before passing the value to mentioned before the init function we replace all characters matched by `replRegexp` with empty string `""`. I found this approach the most flexible and appealing for the user of the decimal library. I gathered opinions from my friends and they also agree with me. 
I may be wrong, but I think it's a much better approach than creating a init function with any other additional arguments such as: `thousandsSeparator` and/or `prefixToRemove`, etc. `replRegex` gives a lot of freedom and flexibility to the users of this library without adding dozens of arguments to the function signature that they probably don't need.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -178,19 +178,19 @@ func NewFromString(value string) (Decimal, error) {
 }
 
 // NewFromFormattedString returns a new Decimal from a formatted string representation.
-// The second argument - replRegexp, is a regexp rule that is used for matching characters occurrences that should
-// be removed from the decimal string representation. All matched characters will be replaced with an empty string.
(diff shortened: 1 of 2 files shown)


mwoss:
I'm merging this PR to master. I feel that Jason would approve this PR :D

njason:
LGTM! Sorry for the late response

PR #192, opened 2020-12-04
Added XML support for NullDecimal

bharath23:
* Added UnmarshalText and MarshalText support for NullDecimal
* Add tests for XML operations on NullDecimal

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -1381,6 +1381,33 @@ func (d NullDecimal) MarshalJSON() ([]byte, error) {
 	return d.Decimal.MarshalJSON()
 }
 
+// UnmarshalText implements the encoding.TextUnmarshaler interface for XML
+// deserialization
+func (d *NullDecimal) UnmarshalText(text []byte) error {
+	str := string(text)
+
+	// check for empty XML or XML without body e.g., <tag></tag>
+	if str == "" {
+		d.Valid = false
+		return nil
+	}
+	if err := d.Decimal.UnmarshalText(text); err != nil {
+		d.Valid = false
+		return err
+	}
+	d.Valid = true
+	return nil
+}
+
+// MarshalText implements the encoding.TextMarshaler interface for XML
+// serialization.
+func (d NullDecimal) MarshalText() (text []byte, err error) {
+	if !d.Valid {
+		return []byte{}, nil
+	}
+	return d.Decimal.MarshalText()
+}
+
 // Trig functions
 
 // Atan returns the arctangent, in radians, of x.
--- a/decimal_test.go
+++ b/decimal_test.go
@@ -766,6 +766,96 @@ func TestBadXML(t *testing.T) {
 	}
 }
 
+func TestNullDecimalXML(t *testing.T) {
+	// test valid values
+	for _, x := range testTable {
+		s := x.short
+		var doc struct {
+			XMLName xml.Name    `xml:"account"`
+			Amount  NullDecimal `xml:"amount"`
+		}
+		docStr := `<account><amount>` + s + `</amount></account>`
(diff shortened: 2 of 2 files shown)


decimal.go
```diff
@@ -1307,6 +1307,31 @@ func (d NullDecimal) MarshalJSON() ([]byte, error) {
...
 
+// UnmarshalText implements the encoding.TextUnmarshaler interface for XML
+// deserialization
+func (d *NullDecimal) UnmarshalText(text []byte) error {
+	str := string(text)
+
+	if str == "" {
+		d.Valid = false
+		return nil
+	}
```
mwoss, line 1318: Is this case necessary? Wouldn't the below if statement be enough?

bharath23: Decimal.Unmarshal returns error. For NullDecimal it is not an error as NullDecimal.Valid will be set to false.

bharath23: It is similar to how NullDecimal.UnmarshalJSON works.

mwoss: Ok, got it :3 Thanks!
I will get back to your PR tomorrow if you don't mind :) I would love to check the unit test too

bharath23: Any updates?

mwoss: I'm so sorry :<. I completely forgot about your pull request Bharath. I'm leaving in about an hour for holiday dinner with family, but I will check those unit test tomorrow.
I'm adding this to my TODO list, so I will not forget about it :D

decimal.go
```diff
@@ -1307,6 +1307,31 @@ func (d NullDecimal) MarshalJSON() ([]byte, error) {
 	return d.Decimal.MarshalJSON()
 }
 
+// UnmarshalText implements the encoding.TextUnmarshaler interface for XML
+// deserialization
+func (d *NullDecimal) UnmarshalText(text []byte) error {
+	str := string(text)
+
+	if str == "" {
```
mwoss, line 1315: So this is basically the case for empty XMLs and XMLs withotu the body e.g `<test></test>`?
Could you add a short comment that would denote that, so it will be easier to refactor in the future if needed? :D

decimal.go
```diff
@@ -1307,6 +1307,31 @@ func (d NullDecimal) MarshalJSON() ([]byte, error) {
...
+		d.Valid = false
+		return nil
+	}
+	if err := d.Decimal.UnmarshalText(text); err != nil {
+		return err
```
mwoss, line 1320: Shouldn't we explicitly set here `d.Valid = false`?

decimal_test.go
```diff
@@ -766,6 +766,75 @@ func TestBadXML(t *testing.T) {
 	}
 }
 
+func TestNullDecimalXML(t *testing.T) {
```
mwoss, line 769: I see that you've based those unit test on NullDecimal JSON tests and there is 3 uni test combined into one. It's not a bad thing in my opinion, but could you separate them by adding a one-liner comment before each section? So next contributors would know what test cases are covered here :3

mwoss (commented):
Overall look good :3 just few comments

decimal.go
bharath23: Done.

decimal.go
bharath23: We are returning error. d is invalid and shouldn't be used by caller. We can set it to false, I dont have strong opinion about it.

decimal_test.go
bharath23: Added comments.

mwoss (approved):
Sorry, I think I've missed an email notification with your PR update. 
Looks good IMO. :3

PR #197, opened 2020-12-11
fix unmarshal binary crash

cw35:


mwoss:
Hi @cw35. Thanks for your contribution. 
Could you present a real-life use case when UnmarshalBinary fails? I was not involved during the implementation of it and I can't think about scenarios when it may fail. 16bit and 8bit exponents?

cw35:
> Hi @cw35. Thanks for your contribution.
> Could you present a real-life use case when UnmarshalBinary fails? I was not involved during the implementation of it and I can't think about scenarios when it may fail. 16bit and 8bit exponents?

Hi. We just encountered some crash while trying to unmarshal bad data. It's user provided data. I think it maybe better if we can check the data and return error here.

decimal.go
```diff
@@ -1022,6 +1022,10 @@ func (d Decimal) MarshalJSON() ([]byte, error) {
 // UnmarshalBinary implements the encoding.BinaryUnmarshaler interface. As a string representation
 // is already used when encoding to text, this method stores that string as []byte
 func (d *Decimal) UnmarshalBinary(data []byte) error {
+	if len(data) < 4 {
+		return fmt.Errorf("error decoding binary %v: too short", data)
```
mwoss, line 1026: It could be only me, but I think an error message should be more verbose. I think we should include a note that data should be at least 5 bytes long for proper unmarshaling. 
What do you think?

decimal_test.go
```diff
@@ -2344,6 +2344,19 @@ func TestBinary(t *testing.T) {
 	}
 }
 
+func TestBadBinary(t *testing.T) {
```
mwoss, line 2347: Also, could we update the test case name to something more verbose too? `BadBinary` is not telling much in my opinion. This test case explicitly checking if input data is long (5 bytes) enough.

PR #202, opened 2021-01-15
Fix incorrect calculation of RoundUp/RoundDown because value after rescaling was ignored

lovung:
# Why 
The current implementation uses the `rescale` method before deciding to Add/Sub 1 for the rescaled value. 
But, it will be wrong in the case the rescaled value equals the original value.

# Fix
Compare the value after rescaling.


mwoss:
Ups, I think I forgot about this case. Thank you for the fix @lovung !
Could you also add a test case to cover that case? :DD

lovung:
@mwoss I have already added 4 test cases. Do you want me to add more? Because I think 4 are enough.

mwoss:
I didn't notice those test cases yesterday, sorry.  :D
Looks good to me :3 One more thing only, could you add a description to this PR with an explanation of what exactly this PR is solving? It would be helpful for potential future code modifications.

mwoss (approved):
Looks good :3 
Only one small request about updating PR description :DD

lovung:
@mwoss I have updated the description. Could you take a look? Thank you so much.

PR #205, opened 2021-01-17
func (d Decimal) MustFloat64()

### LINKED ISSUE #204: func (d Decimal) MustFloat64()
```
// MustFloat64 returns the nearest float64 value for d.
// It doesn't indicate if the returned value is exact.
func (d Decimal) MustFloat64() float64 {
	f, _ := d.Float64()
	return f
}
```

felipeneuwald:
Implementes func (d Decimal) MustFloat64().
Resolves #204.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -804,9 +804,9 @@ func (d Decimal) Float64() (f float64, exact bool) {
 	return d.Rat().Float64()
 }
 
-// MustFloat64 returns the nearest float64 value for d.
+// InexactFloat64 returns the nearest float64 value for d.
 // It doesn't indicate if the returned value represents d exactly.
-func (d Decimal) MustFloat64() float64 {
+func (d Decimal) InexactFloat64() float64 {
 	f, _ := d.Float64()
 	return f
 }
--- a/decimal.go
+++ b/decimal.go
@@ -804,6 +804,13 @@ func (d Decimal) Float64() (f float64, exact bool) {
 	return d.Rat().Float64()
 }
 
+// MustFloat64 returns the nearest float64 value for d.
+// It doesn't indicate if the returned value represents d exactly.
+func (d Decimal) MustFloat64() float64 {
+	f, _ := d.Float64()
+	return f
+}
+
 // String returns the string representation of the decimal
 // with the fixed point.
 //


mwoss:
Thanks for your contribution @felipeneuwald!
I was thinking for a moment about your PR for a bit and I'm not convinced whether the `MustFloat` method should be a part of the decimal library. I'm not sure this wrapper Method would be valuable for this library. Maybe it's just me, not sure :<

Note: Unit tests are missing.

What do you think? @njason

felipeneuwald:
Hey guys, 

There are cases that I don't mind if the returned float64 value is exact or not, and I want to assign the float64 value directly into a struct **using a single line of code**. [See](https://play.golang.org/p/OJt3IKl_Uzj): 

```
package main

import (
	"fmt"

)

type someType struct {
	a float64
	b float64
}

func main() {
	data1 := someType{
		a: float64(1),
		b: decimalToFloat64(decimal.NewFromInt(2)),
	}
	fmt.Printf("%+v\n", data1)

	myDecimal := decimal.NewFromInt(4)
	myFloat, _ := myDecimal.Float64()
	data2 := someType{
		a: float64(3),
		b: myFloat,
	}
	fmt.Printf("%+v\n", data2)

}

// decimalToFloat64 receives a d decimal.Decimal and returns a float64
func decimalToFloat64(d decimal.Decimal) float64 {
	f, _ := d.Float64()
	return f
}
```

In that case, the only way to use a single line of code to assign the value to `data1.b` is by using the function `decimalToFloat64`. A little bit of extra code is needed for `data2.b`. 

There are similar cases in the standard library. See https://golang.org/pkg/html/template/#Must. 

Regarding the unit test, I don't see any case that hasn't been covered in [TestFloat64()](https://github.com/shopspring/decimal/blob/master/decimal_test.go#L273) - but if you guys think this is mandatory to move forward, just let me know, and I'll write the test. 

Cheers, Felipe.

mwoss:
Thanks for showing your point of view on that Felipe!
Let's wait for Jason's opinion, as I want to hear what he thinks about it. :D

twocs:
In general, when I see a function named like MustFloat64, precedence is that it must panic if the underlying function would fail.
e.g. "panics if the error is non-nil" https://golang.org/src/text/template/helper.go


As the result of this PR would be a function that ignores any errors. I'd recommend a name such as:
InexactFloat64

felipeneuwald:
@twocs nice catch. Please check now. Cheers.

mwoss:
Looks good, merging :3

PR #209, opened 2021-02-02
impl: Add mongo decoders and encoders

sirKaskadir:
- added encoder and decoder for official mongo driver
- added go.mod


sirKaskadir:
wrong branch

PR #212, opened 2021-02-04
mod bug fix

mjafaari:


mwoss:
Hi! Could you provide new tests for your change? Current ones were not enough to cover edge cases.

decimal.go
```diff
@@ -610,8 +610,11 @@ func (d Decimal) DivRound(d2 Decimal, precision int32) Decimal {
 
 // Mod returns d % d2.
 func (d Decimal) Mod(d2 Decimal) Decimal {
-	quo := d.Div(d2).Truncate(0)
-	return d.Sub(d2.Mul(quo))
+	res := d
+	for res.Cmp(d2) > 0 {
+		res = res.Sub(d2)
+	}
```
mwoss, line 616: This would fail on basic cases like 7 mod 2, also it does not support negative numbers

mwoss (changes requested):

mwoss:
Issue fixed in #258 and then in #317. Closing.

Issue #178, opened 2020-06-26
Wrong calculate Pow


vxxvvxxv:
I wanted to use your library, but noticed that the function pow does not think correctly. Here is a comparison with math.Pow

```go
fmt.Println(decimal.NewFromFloat(1.15).Pow(decimal.NewFromFloat(12.1666667)))
// result: 5.350250105473711181640625
```

```go
fmt.Println(math.Pow(1.15,12.1666667))
// result: 5.476339866422567
```

mwoss:
Hi @XeHeSi. Yes, we are aware of that, the issue is tracked in #55. Right now it's our main priority to implement is correctly.

PR #213, opened 2021-02-05
Issues With Pow Function Solved & Some Tests Added , Results Checked …

SepehrAsh:
Pow Function Fixed & Some Tests Added.
issue #178 solved

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -614,20 +614,16 @@ func (d Decimal) Mod(d2 Decimal) Decimal {
 	return d.Sub(d2.Mul(quo))
 }
 
-// Pow returns d to the power d2
 func (d Decimal) Pow(d2 Decimal) Decimal {
 	var temp Decimal
-	if d2.IntPart() == 0 {
-		return NewFromFloat(1)
-	}
-	temp = d.Pow(d2.Div(NewFromFloat(2)))
-	if d2.IntPart()%2 == 0 {
-		return temp.Mul(temp)
-	}
-	if d2.IntPart() > 0 {
-		return temp.Mul(temp).Mul(d)
-	}
-	return temp.Mul(temp).Div(d)
+	x := d.value
+	x1 := d.exp
+	y := d2.value
+	k := x.Exp(x, y, nil)
+	a := big.NewInt(int64(x1)).Mul(big.NewInt(int64(x1)), y)
+	temp.value = k
+	temp.exp = int32(a.Int64())
+	return temp
 }
 
 // IsInteger returns true when decimal can be represented as an integer value, otherwise, it returns false.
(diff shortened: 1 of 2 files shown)


decimal.go
```diff
@@ -614,20 +614,16 @@ func (d Decimal) Mod(d2 Decimal) Decimal {
...
-	}
-	return temp.Mul(temp).Div(d)
+	x := d.value
+	x1 := d.exp
+	y := d2.value
+	k := x.Exp(x, y, nil)
+	a := big.NewInt(int64(x1)).Mul(big.NewInt(int64(x1)), y)
+	temp.value = k
+	temp.exp = int32(a.Int64())
+	return temp
```
mwoss, line 626: Personally, I would just return a new instance of Decimal here without using the temp intermediate variable.

decimal.go
```diff
@@ -614,20 +614,16 @@ func (d Decimal) Mod(d2 Decimal) Decimal {
 	return d.Sub(d2.Mul(quo))
 }
 
-// Pow returns d to the power d2
 func (d Decimal) Pow(d2 Decimal) Decimal {
```
mwoss, line 617: Could we add a comment that would explain the given solution, just for future devs who will work on improvements or modification of this method?

decimal_test.go
```diff
@@ -2279,14 +2279,64 @@ func TestDecimal_Cmp2(t *testing.T) {
 }
 
 func TestPow(t *testing.T) {
-	a := New(4, 0)
-	b := New(2, 0)
+	a := NewFromFloat(1.15)
```
mwoss, line 2282: For tests, we are using the "test table" approach,. Could you do the same thing, just to stick to a single convention?

decimal.go
```diff
@@ -614,20 +614,16 @@ func (d Decimal) Mod(d2 Decimal) Decimal {
...
-	}
-	if d2.IntPart() > 0 {
-		return temp.Mul(temp).Mul(d)
-	}
-	return temp.Mul(temp).Div(d)
+	x := d.value
+	x1 := d.exp
+	y := d2.value
+	k := x.Exp(x, y, nil)
+	a := big.NewInt(int64(x1)).Mul(big.NewInt(int64(x1)), y)
```
mwoss, line 623: As it works for examples you gave it would fail when d2 is a decimal with `exp != 0`. For example 1,15 ^ 3,15 produces
```go
d1 := RequireFromString("1.15")
d2 := RequireFromString("3.15")

d1.Pow(d2)
-> 13177096043800968524.66956658119265657462564607005262011901628595176276300244518482010...
```

mwoss (changes requested):
This is a good start, but your implementation only works when `d2` has `exp` property equals 0, otherwise, it produces wrong results :<

Issue #185, opened 2020-10-13
Why there is no π(3.14....) number as constant or function?


MaMrEzO:
Just out of curiosity, I'd looked into your library, and just found one usage of Pi number as follow:
[Pi usage in code `func (d Decimal) satan() Decimal`](https://github.com/shopspring/decimal/blob/ebbf8bbe36744e82feb42bf468dc649d5249bf63/decimal.go#L1347)
But using `NewFromFloat` may not get us actual Pi number as expected as `3.14159265358979323846264338327950288419716939937510582097494459`,
I'd test it and got `3.141592653589793`, But we can do `NewFormString("3.114159265358979323846264338327950288419716939937510582097494459")` then we have `3.14159265358979323846264338327950288419716939937510582097494459` as result.
The function must be:
```go
// satan reduces its argument (known to be positive)
// to the range [0, 0.66] and calls xatan.
func (d Decimal) satan() Decimal {
	Morebits := NewFromFloat(6.123233995736765886130e-17) // pi/2 = PIO2 + Morebits
	Tan3pio8 := NewFromFloat(2.41421356237309504880)      // tan(3*pi/8)
	pi, _ := NewFromString("3.14159265358979323846264338327950288419716939937510582097494459")

	if d.LessThanOrEqual(NewFromFloat(0.66)) {
		return d.xatan()
	}
	if d.GreaterThan(Tan3pio8) {
		return pi.Div(NewFromFloat(2.0)).Sub(NewFromFloat(1.0).Div(d).xatan()).Add(Morebits)
	}
	return pi.Div(NewFromFloat(4.0)).Add((d.Sub(NewFromFloat(1.0)).Div(d.Add(NewFromFloat(1.0)))).xatan()).Add(NewFromFloat(0.5).Mul(Morebits))
}
```

And even a function to provide Pi number:

```go
func PiNumber() Decimal{
  pi, _ := NewFromString("3.14159265358979323846264338327950288419716939937510582097494459")
  return pi
}
```

mwoss:
Hi @MaMrEzO. You're right. There is no reason why there is no such function or constant. Good catch!
I will add this to our backlog. Thank a lot :3

amirphl:
The number **3.141592653589793**23846264338327950288419716939937510582097494459 cannot be fully represented in a 64-bit floating-point format, so it is expected to be approximated as 3.141592653589793.

The inexact representation in binary is: **3.141592653589793**115997963468544185161590576171875.

[...]

PR #214, opened 2021-02-05
PiNumber() implemented (with tests) issue #185

aliash98:


decimal.go
```diff
@@ -1601,3 +1601,9 @@ func (d Decimal) Tan() Decimal {
 	}
 	return y
 }
+
+//PiNumber Returns Pi Number For Calculations
+func PiNumber() Decimal {
+	pi, _ := NewFromString("3.14159265358979323846264338327950288419716939937510582097494459")
+	return pi
+}
```
mwoss, line 1609: As the `PiNumber` function does not provide the capability to set the precision of the number Pi, maybe we should make it a global variable instead of a function?

aliash98: Maybe sending an argument to this function in order to determine the precision wouldn't be a bad idea?

mwoss: I think it would be the best decision here, but it would require more work for sure.

PR #216, opened 2021-02-15
Unmarshal empty JSON string to Zero.

apapsch:
Some systems of unspoken horror may return an empty string when they meant Zero. This change lets the package be more permissive on its input regarding empty JSON string.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -1074,6 +1074,11 @@ func (d *Decimal) UnmarshalJSON(decimalBytes []byte) error {
 		return fmt.Errorf("error decoding string '%s': %s", decimalBytes, err)
 	}
 
+	if str == `""` {
+		*d = NewFromInt(0)
+		return nil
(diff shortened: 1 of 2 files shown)


mwoss:
Personally, I think it is not a way to go. Allowing incorrect JSON to be serialized to the proper value (decimal) is a huge antipattern in my opinion. If someone made a design mistake, this library should not be responsible for fixing that. 
What do you think @njason?

njason:
> Personally, I think it is not a way to go. Allowing incorrect JSON to be serialized to the proper value (decimal) is a huge antipattern in my opinion. If someone made a design mistake, this library should not be responsible for fixing that.
> What do you think @njason?

Agreed, I believe this type of case should be handled in the application's business logic, not the decimal lib.

mwoss:
Closing as members doesn't agree on implementation.

PR #220, opened 2021-03-07
Modify Rounding methods

lovung:
# Why
- The current RoundUp, RoundDown methods seem Ceiling and Floor in Java and many others https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/math/RoundingMode.html
image

# What I did
- Change the name of the current methods
  - RoundUp -> RoundCeil
  - RoundDown -> RoundFloor 
- Add the new method with the names RoundUp and RoundDown.
- Add unit test for the new methods

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -899,7 +899,7 @@ func (d Decimal) Round(places int32) Decimal {
 //     NewFromFloat(545).RoundCeil(-2).String()   // output: "600"
 //     NewFromFloat(500).RoundCeil(-2).String()   // output: "500"
 //     NewFromFloat(1.1001).RoundCeil(2).String() // output: "1.11"
-//     NewFromFloat(-1.454).RoundCeil(1).String() // output: "-1.4"
+//     NewFromFloat(-1.454).RoundCeil(1).String() // output: "-1.5"
 //
 func (d Decimal) RoundCeil(places int32) Decimal {
 	if d.exp >= -places {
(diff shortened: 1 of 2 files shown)


lovung:
@mwoss Please help me to review this PR when you have time. Thank you so much.

mwoss:
I apologize @lovung for the late response. I was aware that you have created a PR, but I'm not feeling good lately, so I mostly sleep after my usual 9-7 work :<. I will try to review it today!

mwoss (approved):
I had the same thoughts when I was implementing RoundUp/Down, but I decided to go with those two rounding methods and in v2 introduce a single, parametrized Round method that will implement all possible rounding methods. 
Anyways, I think adding Floor and Ceil rounding will be valuable and we would eventually add them before reasling v2. 
One thing, I would adjust the docs of RoundUp and RooundDown as now they should be something like:
`RoundUp rounds the decimal away from zero.` and `RoundDown rounds the decimal towards zero.`

lovung:
> in v2 introduce a single, parametrized Round method that will implement all possible rounding methods.

This is a good idea. Do you want me to help to implement it?

mwoss:
I was talking about the v2 after fixing all the current issues, so probably after v1.4. About v2 itself I was thinking about the introduction of NaN, Inf values, removing panics, adding Contex, proper precision handling, etc. So there would be much work to do. 
If you are willing to help with current issues, I would more than happy to see more of your contributions to the decimal library :DD

lovung:
@mwoss I will help if I can. I see some current issues that need to have Context to solve but some are the feature of decimal (don't need to solve).

And one more thing, I want to know when will you merge this PR to the `master` branch?

mwoss:
Fantastic, your help is always welcome. :)
I will merge this PR in a second

Issue #231, opened 2021-06-09
UnmarshalBinary() panics if slice is not expected length.


jmalloc:
I've just encountered a panic when testing invalid input to `UnmarshalBinary()`, due to the fact it access a subslice of `data` without doing any length checks (see https://github.com/shopspring/decimal/blob/master/decimal.go#L1156). 

I suspect it should return an error instead, otherwise the user needs to understand the internal details of the binary format to validate this input before calling `UnmarshalBinary()`.

mwoss:
Hi @jmalloc. Thanks for spotting this bug, this for sure needs to be improved.
#197 should cover this case, unfortunately, the author of this PR did not respond to my review, so I will take over this and try to merge it on the weekend to the master branch.

jmalloc:
Ahh, I didn't notice that other PR, sorry.  I've submitted https://github.com/shopspring/decimal/pull/232 which hopefully addresses your concerns with #197.

mwoss:
No problem. Thanks for submitting the fresh PR :D 
Unfortunately, I probably won't review it till the next weekend as I have only 6 days to finish my master's thesis and I have a lot things to do after 8-5 work. Sorry for that :<

mwoss:
@jmalloc I've merged your PR a moment ago, it looks really great. Thanks a lot for your contribution! :3

jmalloc:
Excellent, thanks! I look forward to a new patch release ;)

PR #232, opened 2021-06-10
Add slice range checks to `UnmarshalBinary()`.

jmalloc:
This PR adds range checks to `Decimal.UnmarshalBinary()`, which prior to this PR would panic if `data` was less than 4 bytes in length.

For symmetry with other methods and this new error I've also added a little extra context to the error returned by `GobDecode()`, to make it clear it occurred within `UnmarshalBinary()`.

This addresses https://github.com/shopspring/decimal/issues/231.


mwoss (approved):
Looks great. Thanks for your contribution! :D

Issue #208, opened 2021-02-01
Too many new(bit.Int),Huge pressure when process GC. how can i avoid that?


lvhuat:
As the title discribed

mwoss:
Hi! Do you mean the `decimal` library is making too many calls to `new(big.Int)` overall or there are too many calls to `new(big.Int)` in your code due to usage of this library? Can you elaborate?

lvhuat:
@mwoss  yes! Not only `new(big.Int)` is call too many times, but also slice field `abs` in `decimal.Decimal`.

mwoss:
Thanks for letting us know. I will look at this in the future, as there are many more important issues than that currently. 
Also, this library is mainly aiming for correctness and precision of calculations rather than speed, so performance issues are resolved at the end.

PR #240, opened 2021-08-08
improve performance of Abs() and Round()

DoubleDi:
This PR improves the performance of Abs() and Round() and is related to #208 .
The goal is to return the original value instead of creating a new one when:
* value is not negative in Abs()
* value already has exact precision in Round()

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -457,6 +457,9 @@ func (d Decimal) rescale(exp int32) Decimal {
 
 // Abs returns the absolute value of the decimal.
 func (d Decimal) Abs() Decimal {
+	if !d.IsNegative() {
+		return d
+	}
 	d.ensureInitialized()
 	d2Value := new(big.Int).Abs(d.value)
 	return Decimal{
@@ -879,6 +882,9 @@ func (d Decimal) StringFixedCash(interval uint8) string {
 // 	   NewFromFloat(545).Round(-1).String() // output: "550"
 //
 func (d Decimal) Round(places int32) Decimal {
+	if d.exp == -places {
+		return d
+	}
 	// truncate to places + 1
 	ret := d.rescale(-places - 1)
 


DoubleDi:
Hi @mwoss! 
Sorry for the ping
Can you please check this and consider merging?

mwoss:
Hi @DoubleDi! Could you provide some benchamarks and approximate

DoubleDi:
@mwoss can send you benchmarks after some time but the improvements are using simple math. 
Current Abs and Round always create a new value and makes allocations event if the value didn't change at all.
That is why i have added some math checks:
1. in Abs if we already have a positive value, then we don't need to negate it and we can just return the original one.
2. in Round if we already have a value with N paces that, equals the rounding parameter, that we do not need to do anything and we should return the original value.

This PR was made, when i saw, that over 20% of CPU in our production service, that was doing heavy calculations, was on the Round function.

mwoss:
I got it, but I was wondering what impact it has on other (typical) use cases. Usually `Abs` are called on negative values only, `Round` also is not used to round decimal to the exact same places after decimal point. 
Those two if-statements should not have much impact on performance, but it's always better to analyze it.

DoubleDi:
For example a typica usecase:
You have unmarshalled a decimal.Decimal for a database or JSON. 
You actually not sure is it positive or not, so you do:
```
v = v.Abs()  // this line creates a new big.Int
```
so if you want to optimize you do
```
if v.IsNegative() {
   v = v.Abs()
}
```
Same logic applies to Round (IsNegative() <=> Exp() == -n) 

Benchmark for passing positive/zero/negative values to Abs()

```
package main

import (
	"testing"

)

var (
	p   = decimal.New(1, 0)
	z   = decimal.Zero
	n   = decimal.New(-1, 0)
	all = []decimal.Decimal{p, z, n}
)

func BenchmarkAbs(b *testing.B) {
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		_ = all[i%3].Abs()
	}
}

func BenchmarkNeg(b *testing.B) {
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		x := all[i%3]
		if x.IsNegative() {
			_ = x.Abs()
		}
	}
}
```
Results:

```
goos: darwin
goarch: amd64
pkg: tmp
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkAbs-16    	24514900	        45.35 ns/op	      37 B/op	       1 allocs/op
BenchmarkNeg-16    	64578939	        18.39 ns/op	      13 B/op	       0 allocs/op
PASS
```

mwoss:
Thanks for the detailed explanation and benchmarks @DoubleDi!
I also checked it for myself, and there is almost no impact on performance by adding those two - if statements if the values are positive (abs) or if the `exp` is different than `places` argument (round).

mwoss (approved):
Looks good :3

mwoss:
Thanks for your contribution @DoubleDi! I'm merging this PR to main branch.

PR #241, opened 2021-08-29
Add method to get original int value

### LINKED ISSUE #237: please add get original value funcion
## in current decimal，IntPart Function to get decimal IntPart value
```go
// IntPart returns the integer component of the decimal.
func (d Decimal) IntPart() int64 {
	scaledD := d.rescale(0)
	return scaledD.value.Int64()
}
```
## BUT, IN SOM TIME, NEED TO GET ORIGINAL VALUE
eg:  **PROTOBUF** , you need define Decimal message, and you need convert between proto.Decimal and decimal.Decimal

so, it's really neet to to get original value function,

To get this purpose, I copied decimal code, and Add **Int64Value** function

**PLEASE ADD**

**PLEASE ADD**

**PLEASE ADD**

```go
// Int64Value returns the original  value.
func (d Decimal) Int64Value() int64 {
	return d.value.Int64()
}
```

##  the follw code is  my proto Decimal define：
```protobuf
// Example: 12345.6789 -> { units = 12345, nanos = 678900000 }
message Decimal {

  // Whole units part of the amount
  int64 val = 1;

  // Nano units of the amount (10^-9)
  // Must be same sign as units
  int32 exp = 2;
}
```

### 1、decimal.Decimal -> proto.Decimal
```go
func FromDecimal(d decimal.Decimal) *proto.Decimal {
	return &Decimal{
		Val: d.Int64Value(),
		Exp: d.Exponent(),
	}
}
```

### 2、proto.Decimal -> decimal.Decimal
```go
func (x proto.Decimal) ToDecimal() decimal.Decimal {
	return decimal.New(x.Val, x.Exp)
}
```

clstb:
Closes #237.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -761,6 +761,14 @@ func (d Decimal) Coefficient() *big.Int {
 	return big.NewInt(0).Set(d.value)
 }
 
+// OriginalIntPart returns the integer component of the decimal
+// scaled by the decimals exponent. In other words, it returns the
+// int64 the decimal was originally constructed with.
+func (d Decimal) OriginalIntPart() int64 {
+	scaledD := d.rescale(d.exp)
+	return scaledD.value.Int64()
+}
+
 // IntPart returns the integer component of the decimal.
 func (d Decimal) IntPart() int64 {
 	scaledD := d.rescale(0)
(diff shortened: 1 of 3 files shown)


mwoss:
Resolved in #244. Closing.

PR #263, opened 2021-11-23
Add MsgPack Marshal and Unmarshal function

piyongcai:
Mote: Marshal max digits is 30. it decimal.IntPart large than 30,it will be lose extra digits.

### DIFF
--- a/decimal_msgpack.go
+++ b/decimal_msgpack.go
@@ -1,0 +1,92 @@
+package decimal
+
+import (
+	"errors"
+)
+
+var (
+	errShortBytes = errors.New("msgp: too few bytes left to read object")
+)
+
+// MarshalMsg implements msgp.Marshaler
+func (d Decimal) MarshalMsg(b []byte) (o []byte, err error) {
+	o = require(b, d.Msgsize())
+	str := d.String()
+	sz := len(str)
+	// limit to 30 digits
+	// note, if d.IntPart size large than 30, will be lose.
+	if sz > 30 {
+		sz = 30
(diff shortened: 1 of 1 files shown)


decimal_msgpack.go
```diff
@@ -0,0 +1,92 @@
...
+
+	o, n := ensure(b, 1+sz)
+	o[n] = byte(0xa0 | sz)
+	n++
+
+	return o[:n+copy(o[n:], str)], nil
+}
+
+// UnmarshalMsg implements msgp.Unmarshaler
+func (d *Decimal) UnmarshalMsg(b []byte) (o []byte, err error) {
```
mwoss, line 36: Could you stick to one type of return value (names/unnamed)? Across the entire library, we use an unnamed one, so I personally opt for this approach.

mwoss (changes requested):
Before diving into the implementation I would like to see a few unit tests. They are handy in case of any refactors, bug catching and implementation changes that could happened in the future

decimal_msgpack.go
piyongcai: already Change MsgPack MarshalMsg and UnmarshalMsg function's return value to unnamed.

piyongcai:
> Before diving into the implementation I would like to see a few unit tests. They are handy in case of any refactors, bug catching and implementation changes that could happened in the future

Test Case was added.

mwoss:
Thanks for providing test cases @piyongcai :) I will try to review your code later this week or next to one, as I need to get familiarized with MsgPack before doing a proper review.

PR #264, opened 2021-11-26
README: Fix URLs of the official Go org pages

sryoya:
Sorry for the PR without an issue 👋 

README uses the old URLs of the Go Playground and Go Doc (They are still valid but redirected to the new ones now.). I updated them as the following.

`godoc.org` -> `pkg.go.dev`
`play.golang.org` -> `go.dev/play`


mwoss:
Hi! :D Sorry for the late reply 🙏 I would love to introduce those doc updates before the 1.4.0 release, could you rebase so I could merge it? :D

Issue #270, opened 2022-01-05
Add support for fmt.GoStringer 


a69:
we  need to print Go-syntax representation of decimal variables.
```
package main
import (
	"fmt"

	"github.com/shopspring/decimal"
)

func main() {
	one := decimal.NewFromInt(1)

	fmt.Printf("%#v\n", one)

	//print: 	decimal.Decimal{value:(*big.Int)(0xc00011e3a0), exp:0}
	//expected:	decimal.NewFromString("1")
}

**Comment by mwoss** [comment]
Hi! Personally, I don't think such change is anyhow valuable for the library. `GoString` method should defines the Go syntax for that value and the default value is more valuable in my opinion than the proposed solution. Returning the `NewFromString` init method in such a case makes no sense. I would suggest forking the library or implementing a custom GoString method for a decimal struct in your own project.
The only meaningful change I can see for GoString method would be printing the value of `decimal.Value` instead of just a memory address for a pointer. Something like this:
```go
// before: 
decimal.Decimal{value:(*big.Int)(0xc00011e3a0), exp:0}

//after (first or second approach)
decimal.Decimal{value:1, exp:0} 
decimal.Decimal{value:1 (*big.Int)(0xc00011e3a0), exp:0}
```

a69:
First, we use GoStringer interface for generate data in go source code for Unit Test from actual values. Right now we have to define GoString, Add, Div and many other method to use our custom decimal struct in our project. I believe this change help people to generate easily Go code with this package. 
Second, your proposal change is not valid go syntax, you set private fields outside of package!
For example you can see the implementation of GoString in time package. https://github.com/golang/go/blob/master/src/time/format.go#L528

mwoss:
That's true if GoString must print valid syntax that can be used outside of the package my proposal is not valid at all, I was not familiar with this interface beforehand.
On the other hand printing value syntax using one of few existing init methods feels odd, it's not generic enough and could be confusing for other users of the library. Initialization via struct literal would be the best here, but not possible as you mentioned earlier.
Additionally the problem with `decimal.RequireFromString(" + d.String() + ")` is that we might lose precision while casting to string, all trailing zeroes are removed.  This is something unacceptable imo.

a69:
I think if `big.Int` has a Method like decimal has `func RequireFromString(value string) Decimal` something like: 
`func (z *Int) RequireSetString(s string, base int) *Int` we could a painless implementation like:
[...]

PR #271, opened 2022-01-05
Add GoString method for Decimal and NullDecimal (#270)

a69:
This PR is for support fmt.GoStringer in #270 issue


treuherz:
I'd find a GoStringer implementation really helpful, for same reasons you describe in #270, but I think it would be more helpful for the method to print `fmt.Sprintf("decimal.New(%d, %d)", d.value, d.exp)` instead of `RequireFromString`. This has the advantage of using the most "natural" constructor in the package, and also will actually get you a string that lets you construct a Decimal with the same behaviour.

For example, under this implementation, you could have
```
d1 := decimal.New(1, 2)
d2 := decimal.New(100, 0)

println(d1.GoString()) // decimal.RequireFromString("100")
println(d2.GoString()) // decimal.RequireFromString("100")
println(d1.Exponent()) // 2
println(d2.Exponent()) // 0
```

I realise `decimal.New(1, 2)` is a little bit harder to read an intuitive value from than `decimal.RequireFromString("100")` but I think it's worth it for more consistent behaviour.

treuherz:
Another possible wrinkle here: the user may not have imported the package as `decimal`. I think the options here are either to leave it, as it'll probably be fine in 99.99% of cases, or remove the `decimal.` from the method, which could be more confusing but at least won't be contradictory. The `fmt` package doesn't give explicit guidance here, although their [example](https://cs.opensource.google/go/go/+/refs/tags/go1.17.5:src/fmt/gostringer_example_test.go;drc=262d4f321a885908def68adc4bed68428722b954) doesn't include the name.

a69:
@treuherz 
`decimal.New` works if d.value always fits in int64.
I think as it says in code comment [decimal.go#L5](https://github.com/shopspring/decimal/blob/master/decimal.go#L5):
`// The best way to create a new Decimal is to use decimal.NewFromString`

To fix this issue we could do something like:
```	s := fmt.Sprintf(`decimal.RequireFromString("%s")`, d.string(false))
	if d.exp > 0 {
		s = s + fmt.Sprintf(".Round(%d)", -d.exp)
	}
	return s```

mwoss:
I've already responded in the linked issue. I think we might try initialization via NewBigInt to not to lose precision.

PR #277, opened 2022-03-11
extend readme to highlight special cases around parsing and printing when using trailing zeros

daemonfire300:


mwoss:
Thanks for your contribution @dreamerjackson and sorry for such a late reply. 
Could we wait a bit, so the author of https://github.com/shopspring/decimal/pull/296 could reply to my comments? This PR may change how we want to extend current documentation. :D

PR #278, opened 2022-03-26
Allocate new(big.Int) in Copy method to deeply clone it

VadimKulagin:
This is necessary to match the description of big.Int: https://pkg.go.dev/math/big#Int
```
  To "copy" an Int value,
  an existing (or newly allocated) Int must be set to
  a new value using the Int.Set method; shallow copies
  of Ints are not supported and may lead to errors.
```

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -418,7 +418,7 @@ func NewFromFloatWithExponent(value float64, exp int32) Decimal {
 func (d Decimal) Copy() Decimal {
 	d.ensureInitialized()
 	return Decimal{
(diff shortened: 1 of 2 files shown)


mwoss:
Thanks for your contribution, that's a great catch! I also apologize for my late response. I wasn't working on this lib for some time, there's a lot of new issues that need to be fixed :D

mwoss:
This PR closes https://github.com/shopspring/decimal/issues/327. Thanks again @VadimKulgin!

Issue #267, opened 2021-12-06
String with thousands separator


mamapi:
How to get string with thousands  separator? 
e.g 4,294,967.45 for  4294967.45

mwoss:
Hi again! 
* If you meant decimal string representation, currently it is not possible as `decimal.String()` is not parametrized. You would have to implement it by yourself.
* If you meant parsing string representation of decimal with thousands separator there is a init method called `decimal.NewFromFormattedString`. It can be used for parsing differently formatted inputs using regular expressions.

Example:
```go
r2 := regexp.MustCompile("[_]")
d2, err := NewFromFormattedString("1_000_000", r2)
```

PR #281, opened 2022-04-11
Add Format()

mcdee:
This adds a `Format()` function to decimal, allowing for setting of a thousands separator (as requested in #267) and decimal separator, as well as the existing functionality for trimming trailing zeros.  This allows for easier use in localized environments.

This replaces the prior internal `string()` function, which could now be removed entirely but that would result in a larger PR so has been left for now.

Worth taking a close look at lines 1032-1034 to ensure they behave as expected, as they alter the flow compared to before this PR.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -1025,6 +1025,73 @@ func (d Decimal) String() string {
 	return d.string(true)
 }
 
+// Format formats a decimal.
+// thousandsSeparator can be empty, in which case the integer value will be displayed without separation.
+// if decimalSeparator is empty and the value is a decimal this will panic.
+func (d Decimal) Format(thousandsSeparator string, decimalSeparator string, trimTrailingZeros bool) string {
+	if d.exp >= 0 {
+		d = d.rescale(0)
+	}
+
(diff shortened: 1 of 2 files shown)


mwoss (commented):
Thanks for the contribution and I deeply apologize for my late review. 🙏 

The ability to format the string representation seems like a nice addition to the library. If I remember correctly, this feature has been already requested in the past.
Sadly, there's one big problem. We cannot use the `Format` as a method name. Go does not support method overloading and there's the [`fmt.Formatter`](https://pkg.go.dev/fmt#Formatter) interface that needs to be implemented on Decimal struct one day. Implementing this interface would allow the comprehensive use of decimals in strings. So I think we should not block it by using this name.

I'm also thinking about whether localized-like Format method should be a part of the Decimal struct. If we would implement `fmt.Formatter`, users could freely use packages like [golang.org/x/text/message](https://pkg.go.dev/golang.org/x/text/message) to format the output to their needs. They offer more advanced formatting options and probably would fulfill all current and future devs needs. What do you think?

PR #296, opened 2022-10-01
Fix #295 | String trim trailing zeros global var

### LINKED ISSUE #295: String Trim Trailing Zeros bool variable
## About

As default `.String()` trunc zeroes:
1.00 -> "1"
2.00 -> "2"

If we set `MarshalJSONWithoutQuotes = false`, in JSON we'll get:
```json
{"decimal": Decimal{1.00}} -> {"decimal": 1}
{"decimal": Decimal{2.00}} -> {"decimal": 2}
{"decimal": Decimal{2.01}} -> {"decimal": 2.01}
```

I think we need to add varibale `StringTrimTrailingZeros`, because in `.String()` true constaint:
```golang
func (d Decimal) String() string {
	return d.string(true)
}
```

Need: 

```golang
func (d Decimal) String() string {
	return d.string(StringTrimTrailingZeros)
}
```

## Motivation
By default, when decimal is output as a string (for example, in JSON), zeros are truncated from it (2.00 -> 2, 3.11 -> 3.11, 13.000 -> 13). But this logic can be changed by this variable.

For example, if you have numeric(10,2) values stored in your database, and you want your API response to always be given 2 decimal places (even 2.00, 3.00, 17.00 [not 2,3,17]), then this variable is a great way out.

vasilbekk:


mcdee:
Also see #281 which provides this plus additional functionality.

davseby:
> Also see #281 which provides this plus additional functionality.

This is different as by globally setting the trim trailing zero bool value you don't need to handle the `JSON` `MarshalText` or `MarshalJSON` methods for structures. The `Format` method would still require you to do so.

vasilbekk:
> > Also see #281 which provides this plus additional functionality.
> 
> This is different as by globally setting the trim trailing zero bool value you don't need to handle the `JSON` `MarshalText` or `MarshalJSON` methods for structures. The `Format` method would still require you to do so.

Absolutely! Thank you for your attention and response, it's really pleasant.

chad-bekmezian-snap (approved):

decimal.go
```diff
@@ -52,6 +52,14 @@ var DivisionPrecision = 16
 // silently lose precision.
 var MarshalJSONWithoutQuotes = false
 
+// StringTrimTrailingZeros should be set to false if you want the decimal stringify without zeros trailing.
+// By default, when decimal is output as a string (for example, in JSON), zeros are truncated from it (2.00 -> 2, 3.11 -> 3.11, 13.000 -> 13).
+// But this logic can be changed by this variable.
+// For example, if you have numeric(10,2) values stored in your database,
+// and you want your API response to always be given 2 decimal places (even 2.00, 3.00, 17.00 [not 2,3,17]),
+// then this variable is a great way out.
```
mwoss, line 60:
```suggestion
// TrimTrailingZeros specifies whether trailing zeroes should be trimmed from a string representation of decimal.
// If set to true, trailing zeroes will be truncated (2.00 -> 2, 3.11 -> 3.11, 13.000 -> 13),
// otherwise trailing zeroes will be preserved (2.00 -> 2.00, 3.11 -> 3.11, 13.000 -> 13.000).
// Setting this value to false can be useful for APIs where exact decimal string representation matters.
```

mwoss: What do you think about it? A bit simplified and a bit more explicit version of your suggestion.

decimal.go
```diff
@@ -52,6 +52,14 @@ var DivisionPrecision = 16
...
+var StringTrimTrailingZeros = true
```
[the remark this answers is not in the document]
mwoss: I also wonder whether we should introduce another config var for specifying NewFromString behaviour. The current implementation does not trim zeroes during the initialization from string, but we saw in the past that some developers would like to see such an option.
Unfortunately, we cannot use this newly introduced variable for both of these cases as we would break backward compatibility.

shuqingzai:
Hope this PR will be completed soon.

JohnAD:
@mwoss @vasilbekk Is there upcoming progress on this? It is a road-blocker for me. A library I'm writing reads a JSON doc with Decimal marshaling. So if it reads

```
{ "foo": 12.300 }
```

and I make no changes to the doc and write it back out again, is it writing

```
{ "foo": 12.3 }
```

For my library, which does git-tracking of JSON, this is a serious problem.

If this has stalled, let me know. I'm happy to clone this branch and make an appended PR with the changes you have suggested. 

While I'm at it I could also add a `Match` method that both compares the "scalar" value (like `Equal` does) as well as the exponent. So, (psuedo-code) `"12.3".Equals("12.30")` but `!"12.3".Match("12.30")`.

In the meantime I'm using the greatcloak fork. But that one has a dependency on the entire MongoDB driver. That is too much overhead for my project.

PR #301, opened 2022-11-24
Fix NumDigits with ensureInitialized

### LINKED ISSUE #292: [Bug] (decimal.Decimal{}).NumDigits == 5
code in decimal
```
func (d Decimal) NumDigits() int {
	// Note(mwoss): It can be optimized, unnecessary cast of big.Int to string
	if d.IsNegative() {
		return len(d.value.String()) - 1
	}
	return len(d.value.String())
}
```
when `d.value` is nil, d.value.String() return `"<nil>"` instead of `"0"`
```
func (x *Int) Text(base int) string {
	if x == nil {
		return "<nil>"
	}
	return string(x.abs.itoa(x.neg, base))
}

// String returns the decimal representation of x as generated by
// x.Text(10).
func (x *Int) String() string {
	return x.Text(10)
}
```

zlasd:
Fix #292 

The library promises

> The zero-value is 0, and is safe to use without initialization

Simply add `ensureInitialized` at the beginning of `NumDigits` to ensure zero-value act as `NewFromInt(0)`.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -815,6 +815,7 @@ func (d Decimal) ExpTaylor(precision int32) (Decimal, error) {
 // NumDigits returns the number of digits of the decimal coefficient (d.Value)
 // Note: Current implementation is extremely slow for large decimals and/or decimals with large fractional part
 func (d Decimal) NumDigits() int {
+	d.ensureInitialized()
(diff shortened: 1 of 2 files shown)


mwoss (approved):
Sorry for the late review! :< Thanks for this fix, I completely forgot about it while implementing this helper function.
LGTM!

PR #305, opened 2022-12-16
Support for the three latest Go version

hiroakis:
I use this module on production a long time. I happened to find the Go versions are old. This PR update Go version on CI and go.mod. And, I also update README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Run `go get github.com/shopspring/decimal`
 
 ## Requirements 
 
-Decimal library requires Go version `>=1.7`
+Decimal library requires Go version `>=1.17`
 
 ## Usage
 
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
 module github.com/shopspring/decimal
 
-go 1.13
+go 1.19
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -1,3 +1,3 @@
 module github.com/shopspring/decimal
 
-go 1.13
+go 1.19
```
mwoss, line 3: Go version in `go.mod` specifies the minimum Go version required to compile a given module, thus I don't think we should change it to 1.19. Decimal can be run with go1.7+

README.md
```diff
@@ -22,7 +22,7 @@ Run `go get github.com/shopspring/decimal`
 
 ## Requirements 
 
-Decimal library requires Go version `>=1.7`
+Decimal library requires Go version `>=1.17`
```
mwoss, line 25: Not true, this library can compied and run with Go1.7+

mwoss:
Those changes don't make much sense. CI has been already adjusted to run on the latest Go version, `go.mod` and documentation changes are unnecessary and would suggest we break compatibility with the older Go version - and we did not.

PR #317, opened 2023-03-31
Fix mod when the divisor is slightly greater than the half of dividend (#312)

### LINKED ISSUE #312: Mod calculation error
With 1.3.1, doing 41 mod 21 yields -1 instead of 20.

zlasd:
`d.DivRound(d2, -d.exp+1)` can round up when the original dividend result is closer to the ceil value. Then the `Truncate` function will not work as expected.

Use `d.QuoRem` with 0 precision can fix this.

Fix #312


spatecon (approved):

mwoss (approved):
Looks great! Thanks for fixing this issue! And deeply apologize for for late review :<

Issue #319, opened 2023-06-11
Add functionality to compare if a decimal value is between two decimals


elirehema:
There should be a way to compare if the decimal value is between two values.

mwoss:
Sounds like a reasonable addition to the library. I will take a look at your PR next week :3 Thanks!

PR #320, opened 2023-06-11
Implement Is between functionality

elirehema:
FIX Add functionality to compare if a decimal value is between two decimals #319

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -4,14 +4,14 @@
 //
 // The best way to create a new Decimal is to use decimal.NewFromString, ex:
 //
-//     n, err := decimal.NewFromString("-123.4567")
-//     n.String() // output: "-123.4567"
+//	n, err := decimal.NewFromString("-123.4567")
+//	n.String() // output: "-123.4567"
 //
 // To use Decimal as part of a struct:
 //
-//     type Struct struct {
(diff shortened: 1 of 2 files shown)


decimal.go
```diff
@@ -900,12 +895,18 @@ func (d Decimal) LessThanOrEqual(d2 Decimal) bool {
 	return cmp == -1 || cmp == 0
 }
 
+// Betwee (LTE) return true when d is between d1 and d2 inclusive
```
mwoss, line 898: Go doc should start with the method name, so I'd suggest:
`IsBetween returns true when d is between (inclusive) d1 and d2, otherwise false.`

decimal_test.go
```diff
@@ -2542,6 +2542,13 @@ func TestDecimal_Equalities(t *testing.T) {
 	if c.LessThanOrEqual(b) {
 		t.Errorf("%q should not be less than or equal %q", a, b)
 	}
+
```
mwoss, line 2545: Could you add a separate test case for `IsBetween` where you verify the correctness of left and right bound inclusiveness?

mwoss (changes requested):
Sorry for the late review and thanks a lot for your contribution!

PR #322, opened 2023-07-13
fix an edge case where overflow check in QuoRem is unreachable

frankli0324:
pretty straightforward, int32 overflows before the overflow check

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -565,7 +565,7 @@ func (d Decimal) QuoRem(d2 Decimal, precision int32) (Decimal, Decimal) {
 		panic("decimal division by 0")
 	}
 	scale := -precision
-	e := int64(d.exp - d2.exp - scale)
+	e := int64(d.exp) - int64(d2.exp) - int64(scale)
 	if e > math.MaxInt32 || e < math.MinInt32 {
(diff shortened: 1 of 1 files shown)


frankli0324:
it's also worth noting that https://github.com/shopspring/decimal/blob/master/decimal_test.go#L2164C1-L2171 didn't include tests for that

decimal.go
```diff
@@ -565,7 +565,7 @@ func (d Decimal) QuoRem(d2 Decimal, precision int32) (Decimal, Decimal) {
 		panic("decimal division by 0")
 	}
 	scale := -precision
-	e := int64(d.exp - d2.exp - scale)
+	e := int64(d.exp) - int64(d2.exp) - int64(scale)
```
mwoss, line 568: Thanks for catching that! :3 LGTM!
Could you add a test case for this edge case? If you don't have time I can work on that.

mwoss (approved):

decimal.go
frankli0324: Sorry I do have no time on this. Could you please handle it?

mwoss: Sure will do, no problem :3

PR #330, opened 2023-10-19
feat: unmarshal empty string as zero

hunjixin:


decimal.go
```diff
@@ -1330,7 +1316,10 @@ func (d *Decimal) UnmarshalJSON(decimalBytes []byte) error {
 	if err != nil {
 		return fmt.Errorf("error decoding string '%s': %s", decimalBytes, err)
 	}
-
+	if len(str) == 0 {
+		*d = Zero
+		return nil
+	}
```
mwoss, line 1322: I'm not sure if this should be the default library behavior. I think in most cases empty string is treated as an incorrect value, we should not implicitly cast it to zero value. I understand there could be use cases for such behavior, but it should be at least hidden behind a config flag.
See https://github.com/shopspring/decimal/pull/216 for additional details.

PR #339, opened 2023-12-28
Add implementation of natural logarithm

mwoss:
This MR adds a natural logarithm function - one missing piece to implement properly working Power implementation. 
MR also adds a `constApproximation` struct that represents mathematical constants like `ln10` with their approximations.

Implementation was heavily inspired by CockroachDB's [apd](https://github.com/cockroachdb/apd/blob/master/context.go#L681) implementation of the natural logarithm. Many thanks to all the great engineers who worked on that library!


mwoss:
GitHub Actions is failing due to this bug - https://github.com/actions/setup-go/issues/169. I need to update libraries used by pipeline and it should resolve the issue.

PR #131, opened 2019-01-29
feat: add uint64 surport

Bububuger:

PR #352, opened 2024-02-13
Add NewFromUint64

agmt:
Adds a new constructor `NewFromUint64(uint64)`, which utilises `big.Int.SetUint64()` existing in Go at least since 1.7 (min supported by `decimal`): https://pkg.go.dev/math/big@go1.7#Int.SetUint64
Improves tests of existing constructors from signed int, covering the minimal negative value.


serprex:
Similar to #131 (but less reformat noise), but the other PR also contains a nice fix for clickhouse

mwoss:
Agree, that fix for Clickhouse would be great to introduce in the 1.4.0 release too :3 I think we could merge this PR and I'll ask @Bububuger for the fix in the new PR, so you both could get contributor credits :D
(Tbh I think the precision parameter in the previous PR is not necessary tho and this PR adds the needed unit test too)

mwoss (approved):
Looks good :3 Thanks for your contribution!

PR #356, opened 2024-03-28
Optimize NumDigits

serprex:
Dividing BitLen by math.Log2(10) is what math/big does underneath

Not including the Int64/Uint64 check makes this slightly slower than old method

Included 2 benchmarks, for 10 digit numbers & 100 digit numbers:

```
-- before
> go test -bench=NumDigit -run=NumDigit
goos: linux
goarch: amd64
pkg: github.com/shopspring/decimal
cpu: AMD Ryzen 7 7840U w/ Radeon  780M Graphics
BenchmarkDecimal_NumDigits10-16     	18317293	        63.87 ns/op
BenchmarkDecimal_NumDigits100-16    	 3645015	       329.6 ns/op
PASS
ok  	github.com/shopspring/decimal	2.779s

-- after
...
BenchmarkDecimal_NumDigits10-16     	143781325	         8.488 ns/op
BenchmarkDecimal_NumDigits100-16    	 5931247	       207.4 ns/op
PASS
ok  	github.com/shopspring/decimal	3.505s
```


mwoss (commented):
Thanks again @serprex for your contribution! :3 Optimizing `NumDigits` was the next task on the roadmap, so I'm really happy that you worked on that :D

Unfortunately, I've spotted a small bug. The new implementation doesn't work for all potential decimals. For example decimal with 500 digits in base and 500 digits in fractional part, instead of returning 1000 it returns 999. By looking at your implementation I think an error correction check should be only applied when `check >= 0`, thus I think we should do something like this:

```go
abs := new(big.Int).Abs(d.value)

estimatedNumDigits := int(float64(abs.BitLen()) / math.Log2(10))

// estimatedNumDigits (lg10) may be off by 1, need to verify
digitsBigInt := big.NewInt(int64(estimatedNumDigits))
errorCorrectionUnit := digitsBigInt.Exp(tenInt, digitsBigInt, nil)

if abs.Cmp(errorCorrectionUnit) >= 0 {
  return estimatedNumDigits + 1
}

return estimatedNumDigits
```

What do you think?

serprex:
Updated with your suggestion. Some fuzzing against previous algo would be ideal, but seems good

decimal.go
```diff
@@ -1224,14 +1224,32 @@ func (d Decimal) Ln(precision int32) (Decimal, error) {
...
-	// Note(mwoss): It can be optimized, unnecessary cast of big.Int to string
-	if d.IsNegative() {
-		return len(d.value.String()) - 1
+	if d.value == nil {
+		return 1
+	}
+
+	if d.value.IsInt64() {
+		i64 := d.value.Int64()
+		if i64 <= (1<<53) && i64 >= -(1<<53) {
```
serprex, line 1234: this range could be extended looking for the first int64 whose inexact float64 doesn't have same number of digits, but decided to keep it simple

mwoss:
> Updated with your suggestion. Some fuzzing against previous algo would be ideal, but seems good. Adding `Cmp` result was a bit too cute it seems

Locally, I've run this function against 100k random decimals and compared it with the old implementation. It seems working fine after adjustments :3 But yea, I agree that we need to introduce a more sophisticated test suite one day. Simple unit testing is not enough if we operate on decimals with high precision.

decimal.go
mwoss: Agree, let's keep it simple. New implementation is already a lot faster than the old one and I'm happy with it :D 
I would only leave a comment for future contributors that 1 << 53 represents the significand precision of floating-point numbers.

decimal.go
```diff
@@ -1224,14 +1224,32 @@ func (d Decimal) Ln(precision int32) (Decimal, error) {
...
+			if i64 == 0 {
+				return 1
+			}
```
mwoss, line 1237: Just for readability, but personally I would move this if-statement one level up. It's not necessary tho. :3

decimal_bench_test.go
```diff
@@ -121,6 +121,34 @@ func BenchmarkDecimal_RoundCash_Five(b *testing.B) {
 	}
 }
 
+func numDigits(b *testing.B, want int, val Decimal) {
+	b.Helper()
+	for i := 0; i < b.N; i++ {
+		if have := val.NumDigits(); have != want {
+			b.Fatalf("\nHave: %q\nWant: %q", have, want)
```
mwoss, line 128: Shouldn't we use the `%d` verb to print integers? :D

decimal_bench_test.go
```diff
@@ -132,7 +160,7 @@ func Benchmark_Cmp(b *testing.B) {
 	}
 }
 
-func Benchmark_decimal_Decimal_Add_different_precision(b *testing.B) {
+func BenchmarkDecimal_Add_different_precision(b *testing.B) {
```
mwoss, line 163: Uu thanks for fixing this :3

mwoss (approved):

decimal.go
serprex: Prefer to leave it, ordering is to minimize comparisons for slow path

decimal_bench_test.go
serprex: Fixed, what I get for copy pasting code

decimal.go
serprex: Added comment

decimal.go
mwoss: Make sense to me, let's leave it as it is :D

mwoss:
After the CI fail, I realized that CI only runs against the `main` branch and is not executed in PRs. I will need to take a look at that.

mwoss:
Oh :< We overlooked one aspect @serprex. `big.Int.IsInt64` and `big.Int.CmpAbs` are not available in Go 1.7 that we are aiming to support. They were introduced in 1.10.

serprex:
@mwoss not sure what constraints determine your minversion, but 1.10 was released 6 years ago & the go compiler works to have backwards compatibility based on go.mod value that I'd be surprised if it's a problem raising minversion

mwoss:
I just wanted to support the oldest Go version possible, this was the main reason. But maybe you're right and I'm exaggerating a bit here. When I started working on this library Go1.7 was only 2 years old, now it's already 6. Rather few people use such an old version of Go, plus what you said about go compiler backwards compatibility.

PR #364, opened 2024-04-09
Support scanning uint64

serprex:
Based on #131 while adding test

Was also going to include #139 but it turns out that PR is incorrect


mwoss:
Thanks for the contribution :D Tbh, I wanted to wait for the original author to rebase their PR and commit those changes, but it has been 5 days and they don't respond in the PR. I will merge your changes and mention both of you in the release notes :3

mwoss (approved):

PR #366, opened 2024-04-23
Implement fmt.Formatter

### LINKED ISSUE #326: Implement fmt.Formatter
It would be nice if `Decimal` types implemented the `fmt.Formatter` interface to allow the use of standard `fmt` verbs when formatting decimals to strings.

For example:
```go
fmt.Printf("%.2f", decimal.RequireFromString("3.14"))
```
Would be equivalent to:
```go
fmt.Printf("%s", decimal.RequireFromString("3.14").StringFixed(2))
```

There are a number of other verbs/flags that could be implemented which can be seen [here](https://pkg.go.dev/fmt#hdr-Printing)

u5surf:
* Fixes #326

We would like to get started implement the fmt.Formatter. At first, we implement the indicators of '%s' and '%f'.

'%s' is due to compatibility for conventional behavior.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -1292,6 +1292,16 @@ func abs(n int32) int32 {
 	return n
 }
 
+func (d Decimal) Format(f fmt.State, verb rune) {
+	switch verb {
+	case 'f':
+		places, _ := f.Precision()
+		fmt.Fprintf(f, d.StringFixed(int32(places)))
+	case 's':
+		fmt.Fprintf(f, d.String())
+	}
+}
+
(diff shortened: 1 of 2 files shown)


decimal.go
```diff
@@ -1292,6 +1292,16 @@ func abs(n int32) int32 {
 	return n
 }
 
+func (d Decimal) Format(f fmt.State, verb rune) {
+	switch verb {
+	case 'f':
+		places, _ := f.Precision()
```
serprex, line 1298: how does this work for plain `%f` without precision specified?

PR #370, opened 2024-05-30
separate trigonometry

langurs:
move trig functions to separate file; move trig constants out of functions


officer47p:
Thanks a lot, mate :+1: 
But here is a catch, I believe that the author of the project was planning to put all the Public functions into the `decimal.go` file... :thinking: 
This structure looks cleaner to me, maybe we could create an internal directory in the project to put all the internally used functions in there and after that, we can only expose `decimal.go`, `trigonometry.go`, or any other category of functions.
But it all depends on the author of the project.

langurs:
I see. I didn't realize that might be their design goal (to put all Public functions in one file).

As for the constants, it's more efficient if they're evaluated once, rather than every time a trig. function is called. Also, I think decimal constants should be created from strings, not from binary floats (but then, some of tests have to be rewritten to match). With a binary float, you may be starting with an inaccurate representation.

Issue #375, opened 2024-07-05
RoundUp() and RoundDown() with Negative Numbers


langurs:
The RoundUp() and RoundDown() functions treat negative numbers the opposite of what I would expect. For example, if you're rounding "up," I would expect the number -1.45 could become -1.4, but not -1.5.

The "Math is Fun" and Wikipedia websites seems to have a good description of the variations (https://www.mathsisfun.com/numbers/rounding-methods.html and https://en.wikipedia.org/wiki/Rounding).

Also, I saw no tests for negative numbers in TestDecimal_RoundAndStringFixed(), so I tried adding some, but got some strange results. You might want to try it.

yzy613:
See #378

PR #378, opened 2024-07-18
feat: RoundHalfUp() and RoundHalfDown()

yzy613:


yzy613:
#375 and https://www.mathsisfun.com/numbers/rounding-methods.html

PR #379, opened 2024-07-23
Increase default precision

EVODelavega:
The addition of the `PowPrecisionNegativeExponent` variable in version 1.4 changes the behaviour of `Pow` significantly. In my use-case, I rely on `Pow` to convert amounts to and from a human-readable format and the same amount in the asset precision. When dealing with crypto (e.g. ETH), 18 decimal places is the de-facto norm.

Having this package default to a seemingly arbitrary precision of 16, especially for negative exponents makes little sense, defaults of 18 means this package can be used without having to rely on messy things like changing global variables to tweak the behaviour in an `init` function


yordis:
@mwoss any opportunity to review this one? I just came across the same situation in the crypto space

PR #387, opened 2024-11-04
Adds MarshalYAML for marshaling yaml

cskeeters:
When marshaling with go-yaml/yaml, numbers are output as double quoted strings.  This adds a custom marshaller (MarshalYAML) that respects a new variable MarshalYAMLWithoutQuotes.

If you decide that you don't want this, I'm happy for it to just live here as a closed PR for anyone else looking for how to do this on their own.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -25,6 +25,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"gopkg.in/yaml.v3"
 )
 
 // DivisionPrecision is the number of decimal places in the result when it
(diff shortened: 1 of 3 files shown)


decimal.go
```diff
@@ -1790,6 +1799,20 @@ func (d Decimal) MarshalJSON() ([]byte, error) {
 	return []byte(str), nil
 }
 
+// MarshalYAML implements the yaml.Marshaler interface.
```
ccoVeille, line 1802: Please try


```suggestion
// MarshalYAML implements the [yaml.Marshaler] interface.
```

I think it's the right syntax for using godoc 1.19 links syntax

decimal.go
```diff
@@ -1790,6 +1799,20 @@ func (d Decimal) MarshalJSON() ([]byte, error) {
...
+func (d Decimal) MarshalYAML() (interface{}, error) {
```
ccoVeille, line 1803:
```suggestion
func (d Decimal) MarshalYAML() (any, error) {
```

decimal.go
```diff
@@ -25,6 +25,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"gopkg.in/yaml.v3"
```
ccoVeille, line 28:
```suggestion

	"gopkg.in/yaml.v3"
```

ccoVeille (approved):
One last thing

PR #389, opened 2024-12-09
WIP: String Trim Trailing Zeros bool variable

JohnAD:
This is a near-duplicate of #296  that preserves the @vasilbeckk2 's work and commits. I will now start making the changes requested in #296.


JohnAD:
I added the following tests for handling trailing zeros on negative precision. Such as "1000 to the nearest 100 should serialize as 1.0E2"

```
{"1.0000E3", "1000.0"}, // 1000 to the nearest tenth
{"1.000E3", "1000"},    // 1000 to the nearest one
{"1.0E3", "1.0E3"},     // 1000 to the nearest hundred
{"1E3", "1E3"},         // 1000 to the nearest thousand
```

(for those not familiar with scientific notation usage, it is required when precision is non-fractional. For example, the population of the U.S. is 314 900 000 or 3.149E8. Any digit after the 9 is an unknown.)

JohnAD:
Sadly, it appears that covering  both cases breaks backwards compatibility. So, I've separated it out. I've added a `AvoidScientificNotation` option variable.

As a bonus, I went ahead and made `d.ScientificNotationString()` a public function and added a unit test.

decimal.go
```diff
@@ -65,6 +65,19 @@ var PowPrecisionNegativeExponent = 16
...
+// otherwise trailing zeroes will be preserved (2.00 -> 2.00, 3.11 -> 3.11, 13.000 -> 13.000).
+// Setting this value to false can be useful for APIs where exact decimal string representation matters.
+var TrimTrailingZeros = true
+
+// AvoidScientificNotation specifies whether scientific notation should be used when decimal is turned
+// into a string that has a "negative" precision.
+//
+// For example, 1200 rounded to the nearest 100 cannot accurately be shown as "1200" because the last two
+// digits are unknown. With this set to false, that number would be expressed as "1.2E3" instead.
+var AvoidScientificNotation = true
```
mwoss: Using negative variable names often leads to double negatives in logic (e.g., if (!AvoidScientificNotation)), which can be confusing. Could we use a positive name like `UseScientificNotation`, `WithScientificNotation`? jut to decrease congnitive load.

What do you think?

mwoss (approved):
Overall looks solid. I won't block merging it. We could fix this variable name later if needed.

Sorry for late review 🙏

PR #130, opened 2019-01-16
Added Sqrt and SqrtRound functionality

ldcicconi:
Based this on https://github.com/shopspring/decimal/pull/111 , but corrected an algorithmic bug. 

Not sure why those other changes are showing up.

PR #111, opened 2018-08-19
Create sqrt.go

kempeng:
Implementation of Square Root for decimals

PR #393, opened 2025-01-08
Add Sqrt and SqrtRound functions

weirdgiraffe:
That is yet another iteration to add `Sqrt` method for decimals. I've already seen previous attempts:

- https://github.com/shopspring/decimal/pull/111
- https://github.com/shopspring/decimal/pull/130

and this solution doesn't reinvent the int and just uses the integer `Sqrt` provided by `big` go package.


weirdgiraffe:
@mwoss I've seen that you've been involved into reviews of #130, so could you maybe take a look at this one as well?

Issue #368, opened 2024-05-13
Race on global variable: factorials


nasermirzaei89:
We get a RACE error on running tests in the production with the flag `-race` enabled.
And the stack shows it comes from here:

I believe that the reason is multiple goroutines are calling method `ExpTaylor` and this method manipulates the shared variable `factorials` (https://github.com/shopspring/decimal/blob/master/decimal.go#L84).
As you used a global variable for caching factorials (https://github.com/shopspring/decimal/pull/229), it's not goroutine safe and should have a lock.

mwoss:
Hi @nasermirzaei89! Thanks for bringing this issue up.
I was aware of this potential race condition problem during the implementation of `ExpTaylor`, but I though my lock-less approach would be sufficient for - https://github.com/shopspring/decimal/blob/master/decimal.go#L1089:L1099. I guess I was wrong, I would need to take a closer look on that 👀

mqzabin:
I'm also having trouble with this race condition.

PR #414, opened 2026-01-26
fix: ExpTaylor race condition

Streppel:
Related: #368

Adds a RWMutex around the shared `factorials` slice.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -25,6 +25,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 )
 
 // DivisionPrecision is the number of decimal places in the result when it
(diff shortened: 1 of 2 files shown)


8xmx8 (commented):
Each call is computed independently; while it may be slightly slower, it is safer and clearer.
```golang
func (d Decimal) ExpTaylor(precision int32) (Decimal, error) {
	var factorials = []Decimal{New(1, 0)}
	// Note(mwoss): Implementation can be optimized by exclusively using big.Int API only
	if d.IsZero() {
		return Decimal{oneInt, 0}.Round(precision), nil
	}
···
```

senago:
Hi! Let's merge and release that fix. I encounter races, and I fear they can lead to pretty bad consequences

mwoss:
@Streppel Thank for the PR! Could you rebase so we could merge it? Overall it looks good!

ArthurMPC:
@Streppel tks for the PR! Lets merge it! lgtm

Streppel:
@mwoss rebased!

mwoss:
Thanks! Merged

PR #416, opened 2026-03-15
chore: rename AvoidScientificNotation to UseScientificNotation

mwoss:
As discussed in https://github.com/shopspring/decimal/pull/389#discussion_r2647304859 we should use positive sounding variable to decrease confusion and cognitive load for end users.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -71,12 +71,12 @@ var MarshalJSONWithoutQuotes = false
 // Setting this value to false can be useful for APIs where exact decimal string representation matters.
 var TrimTrailingZeros = true
 
-// AvoidScientificNotation specifies whether scientific notation should be used when decimal is turned
+// UseScientificNotation specifies whether scientific notation should be used when a decimal is turned
 // into a string that has a "negative" precision.
 //
 // For example, 1200 rounded to the nearest 100 cannot accurately be shown as "1200" because the last two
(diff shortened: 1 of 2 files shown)


decimal.go
```diff
@@ -71,12 +71,12 @@ var MarshalJSONWithoutQuotes = false
...
 
-// AvoidScientificNotation specifies whether scientific notation should be used when decimal is turned
+// UseScientificNotation specifies whether scientific notation should be used when a decimal is turned
 // into a string that has a "negative" precision.
 //
 // For example, 1200 rounded to the nearest 100 cannot accurately be shown as "1200" because the last two
-// digits are unknown. With this set to false, that number would be expressed as "1.2E3" instead.
-var AvoidScientificNotation = true
+// digits are unknown. With this set to true, that number would be expressed as "1.2E3" instead.
+var UseScientificNotation = false
```
ccoVeille, line 79: I'm worried about this PR

A public variable name was renamed, which mean it's a breaking change in the API.

There are code out there that will stop working.

The variable could have been kept with a `//Deprecated: ` tag

I don't think it's a real issue because the last published tag doesn't contain the old variable yet.

But I wanted to share my surprise.

I'm curious about your feedback here

mwoss: Yes, this is a breaking change, but this variable was added between the 1.4.0 release and the next upcoming release. I was fine with doing it just because of that, otherwise I would not touch it under any circumstances

PR #417, opened 2026-04-13
fix: NewFromFloat matches NewFromInt when int64 round-trip is exact

8xmx8:
- Skip roundShortest for integer floats with float64(int64(f))==f to preserve exact decimal value.

- Apply d.neg in big.Int path when nd>=19 (negative large integers).

- Adjust tests and examples for updated string forms; add regression and fuzz tests.

Fixes #411


8xmx8:
Reproducible after modification, Issue 29491

PR #418, opened 2026-04-13
fix issues 411,Reason: NewFromFloat returns the shortest decimal for round-trip.

8xmx8:
Perform an exact decimal expansion strictly based on the true binary floating-point value (rather than the shortest round-trip representation).,see：https://github.com/shopspring/decimal/issues/411


tgkzz (approved):

juanjoseluisgarcia:
why this PR has not been merged yet?

mwoss (commented):
Thanks for opening and PR and helping to fix #411. 

It looks great, however, this changes the public contract of `NewFromFloat` / `NewFromFloat32` from "shortest round-trip representation" to "exact binary value," which is a breaking behavioral change.

Examples:

On master:
```go
NewFromFloat(0.1).String() # "0.1"
NewFromFloat(0.3).String() # "0.3"
```

On this branch:
```go
NewFromFloat(0.1).String() # "0.1000000000000000055511151231257827021181583404541015625"
NewFromFloat(0.3).String() # "0.299999999999999988897769753748434595763683319091796875"
```

Anyone comparing .String() output, persisting, etc, these values will be affected. The "shortest" behavior is the long-standing documented contract and is what most callers may rely on.

So maybe we could fix that narrower case from #411? I meant exact expansion only when the float has no fractional part while preserving shortest round-trip for fractional values? 

If we would like to fix both, I would implement new public method like `NewFromFloatExact`

8xmx8:
My testing was not rigorous enough.

